PR #383, opened 2015-01-23
Interrupt the BodyGenerator thread when writes fail

johngmyers:


http-client/src/main/java/com/proofpoint/http/client/jetty/JettyHttpClient.java
```diff
@@ -604,42 +606,26 @@ public long getLength()
...
                 {
-                    BodyGeneratorOutputStream out = new BodyGeneratorOutputStream(chunks);
                     try {
-                        bodyGenerator.write(out);
-                        out.close();
+                        bodyGeneratorThread.set(Thread.currentThread());
+                        BodyGeneratorOutputStream out = new BodyGeneratorOutputStream(chunks);
+                        try {
+                            bodyGenerator.write(out);
+                            out.close();
```
alecswan: Shouldn't you close this in the finally-clause?

johngmyers: No. Look at `BodyGeneratorOutputStream.close()`.

alecswan: I see, you don't want to update chunks in catch-clause as well as finally-clause. It feels sort of messy that chunks are updated directly in this code and indirectly in BodyGeneratorOutpuStream#close().

johngmyers: It does seem odd that the code gives the BodyGenerator the opportunity to send a DONE chunk. I'm not sure I want to change that in the same pull request.

gwittel:
This change LGTM but I think the overall implementation seems suspect and worth checking out in more detail later.

Issues with the class in general (a larger separate fix to look at later): 

1.  Why not use one of the Jetty supplied ```ContentProvider`` like ```OutputStreamContentProvider``` or ```DefferredContentProvider```.  All I can think of is he wanted to somewhat control the chunks read/written.
2.  The handling of exceptions inside the ```JettyResponseFuture``` seems odd.  I need to do some digging to see the exact behavior.  When an exception is caught in ```JettyResponseFuture#failed```, it gets stored in the future via ```setException```, and the future is marked as completed.  Nothing informs Jetty at this point that the request failed, or downstream in the response listener.  It just gets marked as DONE, and then the exception is not checked later until the response listener completes.

johngmyers:
Rebased, resolved conflicts, squashed, and merged.

PR #385, opened 2015-01-25
Remove race condition inherent to Thread.interrupt()

johngmyers:


http-client/src/main/java/com/proofpoint/http/client/jetty/JettyHttpClient.java
```diff
@@ -723,9 +713,102 @@ protected ByteBuffer computeNext()
...
+                finally {
+                    lock.unlock();
+                }
+            }
+
+            public void put(ByteBuffer e)
+                    throws InterruptedException
+            {
+                checkNotNull(e);
+                final ReentrantLock lock = this.lock;
```
alecswan: Interesting, why create a local variable?

johngmyers: I suspect it's a microoptimization. The code came from `ArrayBlockingQueue`.

gwittel:
LGTM

PR #386, opened 2015-01-28
Create a separate thread pool for running BodyGenerator

johngmyers:
Create a separate thread pool for running BodyGenerator

### DIFF
--- a/http-client/src/main/java/com/proofpoint/http/client/jetty/JettyIoPool.java
+++ b/http-client/src/main/java/com/proofpoint/http/client/jetty/JettyIoPool.java
@@ -25,6 +25,8 @@ public class JettyIoPool
     private final QueuedThreadPool executor;
     private final ByteBufferPool byteBufferPool;
     private final Scheduler scheduler;
+    private final QueuedThreadPool bodyGeneratorExecutor;
+    private final Scheduler bodyGeneratorScheduler;
 
     public JettyIoPool(String name, JettyIoPoolConfig config)
     {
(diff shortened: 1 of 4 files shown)


http-client/src/main/java/com/proofpoint/http/client/jetty/JettyIoPool.java
```diff
@@ -106,11 +116,18 @@ public String toString()
...
     {
         private final ThreadGroup threadGroup;
+        private final Scheduler scheduler;
 
-        private JettyThreadPool(ThreadGroup threadGroup, JettyIoPoolConfig config)
+        private JettyThreadPool(String baseName, int maxThreads, int minThreads)
         {
-            super(config.getMaxThreads(), config.getMinThreads());
-            this.threadGroup = checkNotNull(threadGroup, "threadGroup is null");
+            super(maxThreads, minThreads);
```
gwittel: More of an aside as I don't see this particular error in our systems (yet).

This creates a ```QueuedThreadPool``` with an incoming (blocking) execute queue depth of ```max(minThreads, 8)``` so in the case of:
*  Request thread pool -- 8
*  BodyGenerator thread pool -- 8

Are we OK with this?  Blocking on execute doesn't seem to come into play since ```offer``` is used and a ```RejectedExecutionException``` is thrown when offer fails.

Under heavy load I worry about being able to send in only 8 tasks while waiting for one of [8, 200] (or [1, 200]) threads to pick it up (given all threads might be busy or wedged on slow connections).  I suppose we could later instrument the idle thread count or ```isLowOnThreads``` helper functions to see if this is going to be an issue.

gwittel:
LGTM

johngmyers:
As mentioned in the airlift pull request, this isn't sufficient to solve the problem. I won't be merging this at this time.

We also might consider whether we want a shared `BodyGenerator` pool for all `HttpClient`s.

johngmyers:
Closing pull request without merge. We're going to be deprecating `BodyGenerator` instead.

PR #387, opened 2015-01-28
Rate limit logging exception stacks from BalancingHttpClient

johngmyers:
Rate limit logging exception stacks from BalancingHttpClient


johngmyers:
Fixed commit. Please re-review.

PR #389, opened 2015-01-30
Migrate from BodyGenerator to DynamicBodySource

johngmyers:


http-client/src/main/java/com/proofpoint/http/client/balancing/RetryingResponseHandler.java
```diff
@@ -74,7 +74,7 @@ public Boolean call()
                     exception);
```
gwittel: (lines 72-74).  Missed this last time.  May be good to put  a note that we're intentionally leaving the stack trace out.

johngmyers: That was pull #387

CHANGES
```diff
@@ -1,5 +1,38 @@
...
+  - We changed the reporting client to use DynamicBodySource.
+
+* Events
+
+  - We deprecated EventClient.post(EventGenerator) and EventGenerator.
+    This interface effectively required implementation on top of
+    BodyGenerator. No replacement is currently planned.
+  - We changed the HttpEventClient.post(Iterable) and .post(T...)
+    implementations to use DynamicBodySource.
+
```
gwittel: Missing a note that the reporting client has also been changed to use the ```DynamicBodySource``` as well

gwittel:
LGTM.

Its big enough that its hard to feel like I covered everything well though.

PR #390, opened 2015-01-30
Let clients add injectable providers.

psummers:
Adds multibinder to JaxrsModule for additional injectable provider binders.  JaxrsModule.createResourceConfig() provider now registers the additional provider binders.

Sets priority on TimingFilter DynamicFeature to 100 to make it execute very early.

### DIFF
--- a/bootstrap/pom.xml
+++ b/bootstrap/pom.xml
@@ -8,7 +8,7 @@
     <parent>
         <groupId>com.proofpoint.platform</groupId>
         <artifactId>platform</artifactId>
-        <version>1.13-SNAPSHOT</version>
+        <version>1.13</version>
     </parent>
 
     <dependencies>
--- a/bootstrap/pom.xml
+++ b/bootstrap/pom.xml
@@ -8,7 +8,7 @@
     <parent>
         <groupId>com.proofpoint.platform</groupId>
         <artifactId>platform</artifactId>
-        <version>1.13</version>
+        <version>1.14-SNAPSHOT</version>
     </parent>
 
     <dependencies>
--- a/concurrent/pom.xml
+++ b/concurrent/pom.xml
@@ -8,7 +8,7 @@
     <parent>
         <groupId>com.proofpoint.platform</groupId>
         <artifactId>platform</artifactId>
-        <version>1.13-SNAPSHOT</version>
+        <version>1.13</version>
     </parent>
 
     <dependencies>
--- a/concurrent/pom.xml
+++ b/concurrent/pom.xml
@@ -8,7 +8,7 @@
     <parent>
         <groupId>com.proofpoint.platform</groupId>
         <artifactId>platform</artifactId>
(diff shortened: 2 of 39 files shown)


johngmyers:
Rebased and merged the `TimingFilter` commit.

psummers:
Per discussion:

The current implementation provides a way to let clients add injectable bindings, but the client has to provide an instance of an AbstractBinder to JaxrsBinder.bindInjectableProviderBinderInstance(), for example, to make Principal injectable via @Context:

```Java
@Provider
public class PrincipalInjectableProvider
        implements Factory<Principal>
{
    private final HttpServletRequest request;

    @Inject
    public PrincipalInjectableProvider(HttpServletRequest httpServletRequest)
    {
        this.request = httpServletRequest;
    }

    @Override
    public Principal provide()
    {
        final Object user = request.getAttribute(BasicAuthFilter.AUTHORIZED_USER_KEY);
        return user == null ? null : new Principal()
        {
            @Override
            public String getName()
            {
                return user.toString();
            }
        };
    }

    @Override
    public void dispose(Principal principal)
    {
    }

    public static AbstractBinder principalInjectableProviderBinder()
    {
        return new AbstractBinder()
        {
            @Override
            protected void configure()
            {
                bindFactory(PrincipalInjectableProvider.class).to(Principal.class);
            }
        };
    }
}
```

To bind it:
```Java
jaxrsBinder(binder).bindInjectableProviderBinderInstance(PrincipalInjectableProvider. principalInjectableProviderBinder());
```

I'd much rather have a binder that just associates the provider class with the injected interface, and let platform create the AbstractBinder instance:

```Java
jaxrsBinder(binder).bindInjectableProvider(PrincipalInjectableProvider.class).to(Principal.class);
```
Then then I don't have to implement an AbstractBinder, too.

johngmyers:
A Guice Provider is not permitted to return null.  We should probably have it take a `Provider<Optional<T>>` Can a `Factory` return null?

psummers:
Note that this is a javax.ws.rs.ext.Provider, not a Guice provider.  Factory.provide() is allowed to return null, which is an acceptable value to inject into the resource method.

Did I answer your question, or misunderstand it?

johngmyers:
`javax.ws.rs.ext.Provider` is an annotation, not an interface with a method returning instances of a parameterized type.

johngmyers:
I sent you a (recursive) pull request to improve the API.
The name of the test class needs to be improved. Something like `TestBindContext`.
I'm not sure we want to abuse `Supplier<T>` like that, perhaps we should define our own interface.

johngmyers:
Need to test null context values.

psummers:
That was just to simulate the way it's happening in AuthProvider to satisfy myself that "double-bagging" it wouldn't cause problems.  I'll change it back.

johngmyers:
Replaced by pull #398

PR #395, opened 2015-02-07
Fix findbugs issues

johngmyers:


http-client/src/main/java/com/proofpoint/http/client/Request.java
```diff
@@ -39,11 +39,11 @@
     private final BodySource bodySource;
     private final boolean followRedirects;
 
-    public Request(URI uri, String method, @Nullable ListMultimap<String, String> headers, @Nullable BodySource bodySource)
+    public Request(URI uri, String method, ListMultimap<String, String> headers, @Nullable BodySource bodySource)
```
gwittel: If ```headers``` is no longer nullable, do we need to enforce it with a ```checkNotNull``` or work around with a ```firstNonNull``` given it gets blindly copied?

johngmyers: `ImmutableListMultimap.copyOf()` does that. It apparently never was nullable.

gwittel: ```ImmutableListMultimap.copyOf()``` Doesn't seem to do the check ([1](https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/ImmutableListMultimap.java#L253)):

```java
public static <K, V> ImmutableListMultimap<K, V> copyOf(
      Multimap<? extends K, ? extends V> multimap) {
    if (multimap.isEmpty()) {
      return of();
    }
// ....
```

johngmyers: `multimap.isEmpty()` will throw NPE if `multimap` is null.

gwittel: Right.  If you're fine with the implicit NPE vs the other explicit checks then this is fine.

gwittel:
LGTM.

PR #400, opened 2015-02-18
Allow setting application name from config

johngmyers:
Allow setting application name from config

### DIFF
--- a/bootstrap/src/main/java/com/proofpoint/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/com/proofpoint/bootstrap/Bootstrap.java
@@ -16,6 +16,7 @@
 package com.proofpoint.bootstrap;
 
 import com.google.common.annotations.Beta;
+import com.google.common.base.Function;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableList.Builder;
 import com.google.inject.Binder;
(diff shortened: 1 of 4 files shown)


bootstrap/src/main/java/com/proofpoint/bootstrap/Bootstrap.java
```diff
@@ -77,14 +78,18 @@
...
+        return new StaticBootstrapBeforeModules(applicationName);
     }
 
-    private Bootstrap(String applicationName, Iterable<? extends Module> modules, boolean initializeLogging)
+    public static <T> BootstrapBeforeModules bootstrapApplication(Class<T> configClass, Function<T, String> applicationNameFunction)
+    {
+        return new DynamicBootstrapBeforeModules<>(configClass, applicationNameFunction);
+    }
+
+    private Bootstrap(Module applicationNameModule, Iterable<? extends Module> modules, boolean initializeLogging)
```
gwittel: This seems to just trust that the ```applicationNameModule``` does the right thing and configures an instance of the ```ApplicationNameModule``` somewhere.

Would we prefer to enforce this programmatically vs just letting the service fail during ```NodeInfo``` creation?

johngmyers: That is a private constructor and all extant static factory methods are tested to create a NodeInfo with the right application name.

johngmyers: This commit is obsolete.

gwittel: Agreed; it still trusts whoever touches this constructor does the right thing including tests.  I'm not a fan of this, but given it is internal, I don't feel strongly enough about this to object further.

gwittel:
LGTM

johngmyers:

PR #401, opened 2015-02-19
Reimplement SmileBodyGenerator to not use (deprecated) BodyGenerator

johngmyers:
Reimplement SmileBodyGenerator to not use (deprecated) BodyGenerator


gwittel:
LGTM

PR #403, opened 2015-02-24
Cleanup sample server

johngmyers:
Cleanup sample server

### DIFF
--- a/sample-server/src/main/java/com/proofpoint/platform/sample/StoreConfig.java
+++ b/sample-server/src/main/java/com/proofpoint/platform/sample/StoreConfig.java
@@ -15,13 +15,14 @@
  */
 package com.proofpoint.platform.sample;
 
-import com.google.common.base.Preconditions;
 import com.proofpoint.configuration.Config;
 import com.proofpoint.configuration.LegacyConfig;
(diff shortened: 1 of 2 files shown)


sample-server/src/main/java/com/proofpoint/platform/sample/StoreConfig.java
```diff
@@ -37,9 +38,7 @@ StoreConfig setTtlInMs(int duration)
     @Config("store.ttl")
     StoreConfig setTtl(Duration ttl)
     {
-        Preconditions.checkNotNull(ttl, "ttl must not be null");
-        Preconditions.checkArgument(ttl.toMillis() > 0, "ttl must be > 0");
-
+        checkNotNull(ttl, "ttl must not be null");
```
gwittel: Any particular reason this isn't just done on the getter with a ```@NotNull``` validator?

johngmyers: The default is not null, so there's no legit way for this to be null.

gwittel:
LGTM

johngmyers:
Rebased and merged

PR #410, opened 2015-03-04
Shade asm

johngmyers:


gwittel:
LGTM.

I am trusting the shaded asset list as committed vs digging into whether or not its the complete/necessary set.

johngmyers:
Squashed CHANGES and merged

PR #417, opened 2015-03-19
Upgrade Jersey to 2.17

johngmyers:
Upgrade Jersey to 2.17

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,9 @@
+Platform 1.18
+
+* Library Upgrades
+
+  - Jersey to 2.17
+
 Platform 1.17
 
 * Maven 3.2.3 required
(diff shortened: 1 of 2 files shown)


gwittel:
Jersey and logback LGTM.

Jetty 9.2.11.M0 -- There is a follow up fix to the original connection leak bugfix that came after M0 was tagged (eclipse/jetty.project@07d29fb1b39740b298ba2497ec1b06e86c1e7efc.).  Do you want to wait for this to go mainline?

johngmyers:
Merged Jersey and logback.

Upon cursory inspection, the followup fix appears to be an unusual corner case. A 100-continue response that causes the request to complete.

For what it's worth, Airlift has released with 9.2.11.M0. I could go either way with this one.

johngmyers:
Jetty 9.3.0 will be coming out in a few weeks, so we'll just wait for that.

PR #418, opened 2015-03-28
Remove the "self" pointer from the sample server's protocol

johngmyers:
Remove the "self" pointer from the sample server's protocol


gwittel:
LGTM

johngmyers:
Squashed and merged

PR #427, opened 2015-05-14
Update Jetty client timeout config

johngmyers:


gwittel:
Code LGTM.

Missing docs on configuration parameter name change/deprecation.

PR #431, opened 2015-05-20
Create /admin/stop-announcing resource to admin port

johngmyers:
Create /admin/stop-announcing resource to admin port

### DIFF
--- a/jmx-http/src/main/java/com/proofpoint/jmx/StopAnnouncingResource.java
+++ b/jmx-http/src/main/java/com/proofpoint/jmx/StopAnnouncingResource.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2010 Proofpoint, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


jmx-http/src/main/java/com/proofpoint/jmx/StopAnnouncingResource.java
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright 2010 Proofpoint, Inc.
```
gwittel: 2015

jmx-http/src/main/java/com/proofpoint/jmx/StopAnnouncingResource.java
```diff
@@ -0,0 +1,50 @@
...
+        this.adminServerCredentialVerifier = requireNonNull(adminServerCredentialVerifier, "adminServerCredentialVerifier is null");
+    }
+
+    @PUT
+    public void stopAnnouncing(@HeaderParam("Authorization") String authHeader)
+    {
+        adminServerCredentialVerifier.authenticate(authHeader);
+
+        log.info("Received shutdown request. Stopping discovery announcer.");
+        announcer.destroy();
```
gwittel: Do we care if multiple shutdown requests come in while one is in progress?  Seems like the Announcer won't care; (worst case, multiple de-announces)

johngmyers: Don't think so. Multiple de-announces are fine, as that is an idempotent operation.

gwittel:
LGTM

johngmyers:
Rebased, squashed, and merged.

PR #437, opened 2015-05-22
Disable DHE ciphers

johngmyers:
Disable DHE ciphers


gwittel:
LGTM

We should put a short note about the impact (no support for IE6/XP, no forward secrecy on clients that don't support ECDH)?

PR #440, opened 2015-06-02
Destroy Jetty client connections stuck in closing

johngmyers:
Destroy Jetty client connections stuck in closing

Jetty client connections can sometimes get stuck while closing which reduces
the available connections.  The Jetty Sweeper periodically scans the active
connection pool looking for connections in the closed state, and if a connection
is observed in the closed state multiple times, it logs, and destroys the connection.


gwittel:
LGTM

PR #446, opened 2015-06-17
Create HttpClientBinder.bindBalancingHttpClient() methods

johngmyers:


http-client/src/main/java/com/proofpoint/http/client/StaticHttpServiceBalancerProvider.java
```diff
@@ -0,0 +1,68 @@
...
+import static java.util.Objects.requireNonNull;
+
+class StaticHttpServiceBalancerProvider implements Provider<HttpServiceBalancer>
+{
+    private final String type;
+    private final Set<URI> baseUris;
+    private ReportCollectionFactory reportCollectionFactory;
+
+    StaticHttpServiceBalancerProvider(String type, Set<URI> baseUris)
+    {
```
gwittel: Do we want to reject construction of a static balancer with 0 URIs?  Or keep the same behavior as discovered services with 0 members -- where it will fail upon the first attempt?

johngmyers: Addressed in subsequent commit.

http-client/src/main/java/com/proofpoint/http/client/HttpClientBinder.java
```diff
@@ -55,6 +68,41 @@ public HttpClientBindingBuilder bindHttpClient(String name, Class<? extends Anno
         return createBindingBuilder(new HttpClientModule(name, annotation, rootBinder));
     }
 
+    public BalancingHttpClientBindingBuilder bindBalancingHttpClient(String name, Class<? extends Annotation> annotation, Set<URI> baseUris)
+    {
+        checkNotNull(name, "name is null");
+        checkNotNull(annotation, "annotation is null");
+        checkNotNull(baseUris, "baseUris is null");
+        checkArgument(!baseUris.isEmpty(), "baseUris is empty");
```
gwittel: Do we want the not empty check in the ``StaticHttpServiceBalancerProvider``  constructor instead?

johngmyers: `StaticHttpServiceBalancerProvider` is package local, so it doesn't really matter. Sooner seems better.

gwittel:
LGTM

johngmyers:
Rebased and merged

PR #456, opened 2015-09-14
Add HttpServer.BusyThreads.Max metric

johngmyers:
Add HttpServer.BusyThreads.Max metric

### DIFF
--- a/http-server/src/main/java/com/proofpoint/http/server/HttpServer.java
+++ b/http-server/src/main/java/com/proofpoint/http/server/HttpServer.java
@@ -21,6 +21,7 @@ import com.google.common.primitives.Ints;
 import com.proofpoint.bootstrap.AcceptRequests;
 import com.proofpoint.http.server.HttpServerBinder.HttpResourceBinding;
 import com.proofpoint.node.NodeInfo;
+import com.proofpoint.stats.MaxGauge;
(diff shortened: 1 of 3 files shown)


http-server/src/main/java/com/proofpoint/http/server/HttpServer.java
```diff
@@ -119,7 +122,16 @@ public HttpServer(HttpServerInfo httpServerInfo,
         checkNotNull(queryStringFilter, "queryStringFilter is null");
         checkNotNull(theServlet, "theServlet is null");
 
-        QueuedThreadPool threadPool = new QueuedThreadPool(config.getMaxThreads());
+        QueuedThreadPool threadPool = new QueuedThreadPool(config.getMaxThreads()) {
+
+            @Override
+            protected void runJob(Runnable job)
```
gwittel: Is there a chance that the +1 and -1 will get split across buckets?  Could that make the min/max look weird under certain conditions?

johngmyers: That adjusts the current value of `MaxGauge`. Only the maximums are bucketed.

gwittel: Right; makes sense.

gwittel:
LGTM

PR #469, opened 2015-11-04
Specify defaults for https keystore key/path

johngmyers:


gwittel:
LGTM.  Though I worry about people leaving the password as default out of laziness.

johngmyers:
With our deployment method, that password isn't actually security sensitive.

PR #477, opened 2015-11-20
Jetty 9.3

johngmyers:


gwittel:
LGTM.

We should document the HTTP client behavior with respect to when max requests queued per destination is zero (vs non-zero).

johngmyers:
Added an item to CHANGES

PR #478, opened 2015-12-08
Skeleton server runs HTTPS by default

johngmyers:
Skeleton server runs HTTPS by default

### DIFF
--- a/skeleton-server/etc/config.properties
+++ b/skeleton-server/etc/config.properties
@@ -1,2 +1,4 @@
 node.environment=test
 testing.discovery.uri=http://localhost:4111
+http-server.http.enabled=true
+http-server.https.enabled=false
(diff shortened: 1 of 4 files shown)


skeleton-server/etc/config.properties
```diff
@@ -1,2 +1,4 @@
 node.environment=test
 testing.discovery.uri=http://localhost:4111
+http-server.http.enabled=true
```
gwittel: Did we want these to match the application defaults?

johngmyers: No, we only want to put things in config.properties that don't match the application defaults.

The idea is to allow testing in the IDE without installing a keystore.

gwittel:
LGTM

PR #492, opened 2016-03-05
Log response body when reporting service returns error

johngmyers:
Log response body when reporting service returns error


gwittel:
LGTM

johngmyers:
Rebased and merged

PR #513, opened 2016-07-22
Upgrade Jetty to 9.3.10

johngmyers:


alinav:
LGTM http://central.maven.org/maven2/org/eclipse/jetty/jetty-project/9.3.11.M0/jetty-project-9.3.11.M0-version.txt we might want to pick that one

johngmyers:
I don't think we want a milestone build

PR #524, opened 2016-08-22
Merge selected airlift changes, round three

johngmyers:


gwittel:
LGTM.

I may create a request later to expose hooks on configuring the ```GzipHandler``` (in particular exclude paths, mime types).  We've run into some cases where the handler causes trouble by buffering streamed output vs sending to clients immediately.

johngmyers:
Rebased and merged

PR #526, opened 2016-08-24
Re-enable http-server support for the SSLv2Hello protocol

johngmyers:
Re-enable http-server support for the SSLv2Hello protocol


dmitry1975:
Looks good.

PR #546, opened 2016-10-25
Disable JAX-RS resolution of relative URIs in Location

johngmyers:


alinav:
LGTM, but I would prefer a generic solution in next iteration

PR #560, opened 2016-12-05
HttpClient forward trace tokens by default

johngmyers:
HttpClient forward trace tokens by default


johngmyers:
Rebased and merged

PR #563, opened 2016-12-08
Upgrade dependencies

johngmyers:


library/pom.xml
```diff
@@ -1000,7 +1000,7 @@
             <dependency>
                 <groupId>org.mockito</groupId>
                 <artifactId>mockito-core</artifactId>
-                <version>1.10.8</version>
+                <version>2.2.28</version>
```
gwittel: In the release notes, it would be good to note that Mockito 2.x has breaking changes vs 1.x so there's a clear pointer to the fixes that may be involved:

https://github.com/mockito/mockito/wiki/What's-new-in-Mockito-2#incompatible

johngmyers: Most if not all of those incompatibilities are unlikely to affect us. Possibly the anyX() / any() one, but the rest either aren't relevant or are for advanced use.

Which ones do you think merit mentioning?

gwittel: `anyX` and Hamcrest style `argThat()` family.  I know there are places that use the hamcrest style matches already.

johngmyers:
Added CHANGES comment and upgraded logback to pull in the removal of a fill-the-disk misfeature.

johngmyers:
Rebased, squashed fixups, and merged

PR #564, opened 2016-12-10
Allow trace tokens to carry additional properties

johngmyers:


johngmyers:
Doesn't add JSON support to the event client. It didn't seem worthwhile to do that work.

http-server/src/main/java/com/proofpoint/http/server/TraceTokenFilter.java
```diff
@@ -45,11 +52,19 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
...
-            registerRequestToken(token);
+        if (token == null || token.isEmpty()) {
+            createAndRegisterNewRequestToken();
+        }
+        else if (token.charAt(0) == '{') {
+            try {
+                registerTraceToken(TRACE_TOKEN_JSON_CODEC.fromJson(token));
+            }
+            catch (RuntimeException e) {
+                createAndRegisterNewRequestToken();
```
ttockpfpt: I would change this. If the caller passes in a trace token that starts with {, but it doesn't parse as json, then I would just call registerRequestToken() using the raw string value. That way we continue to pass along the trace token that was passed in to us.  Creating a new request token when we were given one doesn't seem right.

http-server/src/main/java/com/proofpoint/http/server/TraceTokenFilter.java
```diff
@@ -45,11 +52,19 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
...
+            }
         }
         else {
-            createAndRegisterNewRequestToken();
+            registerRequestToken(token);
```
ttockpfpt: When does this get unset?  Or do we just rely on the next call in to this method to replace the thread-local-variable for this thread?

trace-token/src/main/java/com/proofpoint/tracetoken/TraceTokenManager.java
```diff
@@ -67,35 +95,54 @@ public static TraceTokenScope registerRequestToken(@Nullable String token)
...
 
     /**
-     * Create and register a new trace token.
-     * @return The created token.
+     * Create and register a new trace token with a random id.
+     * @param properties Additional properties to include in the token.
+     * @return The id of the created token.
      */
-    public static String createAndRegisterNewRequestToken()
+    public static String createAndRegisterNewRequestToken(String... properties)
```
johngmyers: Backwards compatibility prevents us from returning the scope. If you need a scope, this call is not appropriate—call registerTraceToken(null).

trace-token/src/main/java/com/proofpoint/tracetoken/TraceTokenManager.java
```diff
@@ -112,10 +159,34 @@ public static void clearRequestToken()
...
+     * Add properties to the current thread's trace token.
+     *
+     * @param properties Properties to add or replace.
+     * @return a {@link TraceTokenScope} which may be used to restore the thread's
+     * previous set of properties.
+     */
+    public static TraceTokenScope addTraceTokenProperties(String... properties)
+    {
+        TokenState tokenState = token.get();
+        checkState(tokenState != null, "no current trace token");
```
johngmyers: I believe this call is only appropriate in the case where it is known that there is a current token. When would it be appropriate to call this when it isn't known that there is a request being processed?

http-server/src/main/java/com/proofpoint/http/server/TraceTokenFilter.java
johngmyers: We currently rely on the next call to replace the token. It is an assumption that the http-server threads don't do anything but process requests that go through this filter.

trace-token/src/main/java/com/proofpoint/tracetoken/TraceTokenManager.java
```diff
@@ -67,35 +95,54 @@ public static TraceTokenScope registerRequestToken(@Nullable String token)
...
+     * Create and register a new trace token with a random id.
+     * @param properties Additional properties to include in the token.
+     * @return The id of the created token.
      */
-    public static String createAndRegisterNewRequestToken()
+    public static String createAndRegisterNewRequestToken(String... properties)
     {
         String newToken = UUID.randomUUID().toString();
         registerRequestToken(newToken);
+        if (properties.length != 0) {
```
johngmyers: addressed in subsequent commit

http-client/src/main/java/com/proofpoint/http/client/jetty/JettyHttpClient.java
```diff
@@ -1147,7 +1148,7 @@ protected ByteBuffer computeNext()
             {
                 ByteBuffer chunk = chunks.poll();
                 while (chunk == null) {
-                    try (TraceTokenScope ignored = registerRequestToken(traceToken)) {
+                    try (TraceTokenScope ignored = registerTraceToken(traceToken)) {
```
johngmyers: addressed in subsequent commit

http-server/src/main/java/com/proofpoint/http/server/TraceTokenFilter.java
```diff
@@ -53,14 +68,14 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
 
         String token = request.getHeader("X-Proofpoint-TraceToken");
         if (token == null || token.isEmpty()) {
```
alinav: nitpik: Strings.isNullOrEmpty()

trace-token/src/main/java/com/proofpoint/tracetoken/TraceTokenManager.java
```diff
@@ -138,6 +138,7 @@ public static TraceToken getCurrentTraceToken()
      */
     public static String createAndRegisterNewRequestToken(String... properties)
     {
+        checkArgument((properties.length % 2) == 0, "odd number of elements in properties");
```
alinav: this is not backward compatible change. If we used to have string passed in, it will break by forcing it to be in new format. Please create new API for name-value pair thing and keep the other one in.

johngmyers: The original signature had no parameters, so if you used to have a string passed in it wouldn't compile.

trace-token/src/main/java/com/proofpoint/tracetoken/TraceTokenManager.java
```diff
@@ -34,28 +40,50 @@ private TraceTokenManager()
...
+        TokenState oldTokenState = TraceTokenManager.token.get();
 
         String oldThreadName;
-        if (oldToken == null) {
+        if (oldTokenState == null) {
             oldThreadName = currentThread().getName();
         }
         else {
-            oldThreadName = oldToken.getOldThreadName();
+            oldThreadName = oldTokenState.getOldThreadName();
```
ttockpfpt: This logic might lead to the thread name getting reset to something different from what it is currently.  For example:
```
  try (TraceTokenScope ignored = registerTraceToken("bar")) {
    Thread.currentThread().setName("something");
    try (TraceTokenScope ignored2 = registerTraceToken("foo")) {
    }
    // Here the thread name is no longer "something". 
  }
```
I would expect the close() method of the TraceTokenScope to restore things back to the way they were.  Instead it gives me the thread name as saved from the previous trace token.

johngmyers:
Rebased, squashed review fixes, and merged

PR #568, opened 2016-12-21
Support AutoValue event classes

johngmyers:


gwittel:
Looks good overall.  Here we no longer cover non-AutoValue event classes explicitly.  Do we feel the need for exlicit coverage, or is the generated Autovalue code equivalent enough?

johngmyers:
There are several test cases that use non-AutoValue event classes. They are primarily cases that AutoValue doesn't cover, like event classes that subclass other event classes.

gwittel:
Ah ok, I was only looking at the diff vs the full codebase.

johngmyers:

PR #569, opened 2016-12-22
Upgrade Jetty and work around infinite-blocking bug

johngmyers:


http-server/src/main/java/com/proofpoint/http/server/HttpServer.java
```diff
@@ -164,6 +164,7 @@ protected void runJob(Runnable job)
             HttpConfiguration httpConfiguration = new HttpConfiguration();
             httpConfiguration.setSendServerVersion(false);
             httpConfiguration.setSendXPoweredBy(false);
+            httpConfiguration.setBlockingTimeout(config.getNetworkMaxIdleTime().toMillis() + 500);
```
glent: According to the documentation, 
> This timeout is in addition to the Connector.getIdleTimeout(), and applies to the total operation (as opposed to the idle timeout that applies to the time no data is being sent).

So this probably results in a blocking timeout of config.getNetworkMaxIdleTime() * 2 + 500

glent:
LGTM

johngmyers:
Squashed fixup commit and merged

PR #571, opened 2016-12-26
Set default exception handler in Bootstrap when initializing logging

johngmyers:
Set default exception handler in Bootstrap when initializing logging


johngmyers:

PR #574, opened 2017-01-11
Stop including trace token twice in the launcher.log

johngmyers:


johngmyers:
Holding off merging this until various internal tools are updated.

johngmyers:
Rebased and merged

PR #575, opened 2017-01-11
Support @Key annotations on JAX-RS resource method parameters

johngmyers:


jaxrs/src/main/java/com/proofpoint/jaxrs/TimingResourceDynamicFeature.java
```diff
@@ -100,4 +129,68 @@ else if (type.isAnnotationPresent(Path.class)) {
...
+
+                            synchronized (registeredMap) {
+                                SparseTimeStat existingStat = registeredMap.get(key);
+                                if (existingStat != null) {
+                                    reinsertedSet.add(existingStat);
+                                    return existingStat;
+                                }
+                                registeredMap.put(key, returnValue);
+                                Builder<String, String> tagBuilder = ImmutableMap.builder();
+                                for (int i = 0; i < keyNames.size(); ++i) {
```
gwittel: General clarification -- Am I correct in reading this that the somewhat opaque detail is that `keyNames` and `key` are both derived from the data created by the bytebuddy code generation?  
Therefore the assumption here that `keyNames` and `key` will always have same length and ordering.

This behavior relies on code in `TimingFilter` generating the tag/cache lookups the same way as the `TimingWrapper` generates the data used to populate the loadingcache.

Do we need tests to cover the scenario where only one side is updated (say a new default metric) causing a length mismatch?

johngmyers: `CacheImplementation` came in the refactor prior to adding ByteBuddy.

But there is an invariant that the keyNames produced by `TimingResourceDynamicFeature.configure(...)` to create the `LoadingCache` match in number and order the values given to the `LoadingCache` by `TimingFilter.filter(...)`.

I don't see a useful test for this particular point. It is an implementation detail.

johngmyers: To clarify, the existing tests of the API, which check that particular metrics are produced for particular calls, would catch such a mismatch.

johngmyers:
Rebased and merged

PR #579, opened 2017-01-31
Event client support for new-style trace tokens

johngmyers:


event/src/main/java/com/proofpoint/event/client/EventJsonSerializer.java
```diff
@@ -16,29 +16,45 @@
...
-    public EventJsonSerializer(NodeInfo nodeInfo, @Nullable String token, EventTypeMetadata<T> eventTypeMetadata)
+    EventJsonSerializer(NodeInfo nodeInfo, @Nullable TraceToken token, EventTypeMetadata<T> eventTypeMetadata)
     {
-        this.token = token;
-        this.eventTypeMetadata = checkNotNull(eventTypeMetadata, "eventTypeMetadata is null");
+        if (token == null) {
+            this.token = null;
+        }
+        else if (token.size() == 1) {
+            this.token = token.get("id");
```
ttockpfpt: This line means the trace token that is passed through might not be what the caller provided.

event/src/main/java/com/proofpoint/event/client/EventJsonSerializer.java
```diff
@@ -16,29 +16,45 @@
...
+        }
+        else {
+            try {
+                this.token = new ObjectMapperProvider().get().writeValueAsString(token);
```
ttockpfpt: It's expensive to create an ObjectMapper, should use JsonCodec.OBJECT_MAPPER_SUPPLIER.  And BTW, it would be handy if non-platform code could use the OBJECT_MAPPER_SUPPLIER as well so that application code doesn't have to create its own copy of an ObjectMapper.

event/src/main/java/com/proofpoint/event/client/EventTypeMetadata.java
```diff
@@ -149,8 +150,10 @@ else if (isMultimap(dataType)) {
             if (eventField.fieldMapping() != EventFieldMapping.DATA) {
                 // validate special fields
                 if (containerType != null) {
-                    addMethodError("non-DATA fieldMapping (%s) not allowed for %s", method, eventField.fieldMapping(), containerType);
-                    continue;
+                    if (eventField.fieldMapping() == EventFieldMapping.TRACE_TOKEN && !method.getReturnType().equals(TraceToken.class)) {
```
ttockpfpt: Why not allow a String return type as well?  I think platform should allow trace tokens to flow through without interpretation if that's what the application wants.

event/src/main/java/com/proofpoint/event/client/JsonEventSerializer.java
```diff
@@ -69,8 +72,33 @@ public JsonEventSerializer(NodeInfo nodeInfo, Class<?>... eventClasses)
...
+    public <T> void serialize(T event, @Nullable String token, JsonGenerator jsonGenerator)
+            throws IOException
+    {
+        requireNonNull(event, "event is null");
+        requireNonNull(jsonGenerator, "jsonGenerator is null");
+
+        TraceToken traceToken = null;
+        if (token != null) {
+            try (TraceTokenScope scope = registerRequestToken(token)) {
+                traceToken = getCurrentTraceToken();
```
ttockpfpt: just call serialize(event, getCurrentTraceToken(), jsonGenerator) here to avoid duplicated code between the two flavors of serialize().

event/src/main/java/com/proofpoint/event/client/JsonEventSerializer.java
```diff
@@ -69,8 +72,33 @@ public JsonEventSerializer(NodeInfo nodeInfo, Class<?>... eventClasses)
...
+    @Deprecated
+    public <T> void serialize(T event, @Nullable String token, JsonGenerator jsonGenerator)
+            throws IOException
+    {
+        requireNonNull(event, "event is null");
+        requireNonNull(jsonGenerator, "jsonGenerator is null");
+
+        TraceToken traceToken = null;
+        if (token != null) {
+            try (TraceTokenScope scope = registerRequestToken(token)) {
```
ttockpfpt: This has the potential side effect of changing the current thread's name, which would be very surprising to a caller that is just trying to marshal an event.

event/src/main/java/com/proofpoint/event/client/EventJsonSerializer.java
johngmyers: It is what the caller provided. `TraceToken` is guaranteed to have an `id` field. If there are no additional properties, this causes it to be serialized like it was before properties were introduced. This is consistent with `TraceToken` serialization in HTTP headers.

event/src/main/java/com/proofpoint/event/client/EventJsonSerializer.java
johngmyers: Okay, we can memoize the `ObjectMapper`. We don't want to turn on INDENT_OUTPUT, though.

There is a problem with `JsonCodec` in that it ignores any serializers, etc. that get added through Guice. Not relevant for this set of changes, though.

event/src/main/java/com/proofpoint/event/client/EventTypeMetadata.java
johngmyers: String is permitted, as proven by the `AbstractTestJsonEventWriter.testFixedStringToken()` unit test.

event/src/main/java/com/proofpoint/event/client/JsonEventSerializer.java
johngmyers: It uses a `TraceTokenScope`, so the old thread name is put back.

It is highly unlikely anything is calling this now-deprecated method.

event/src/main/java/com/proofpoint/event/client/JsonEventSerializer.java
johngmyers: Okay

event/src/main/java/com/proofpoint/event/client/EventJsonSerializer.java
johngmyers: On second thought, I'm replacing the ObjectMapper with a JsonCodec

event/src/main/java/com/proofpoint/event/client/JsonEventSerializer.java
ttockpfpt: Good point about it probably not being used.  But thread name is not necessarily restored properly by registreRequestToken(), if there is a trace token but the caller has changed the thread name to something else.

event/src/main/java/com/proofpoint/event/client/EventJsonSerializer.java
ttockpfpt: TraceToken tries to guarantee that there is an "id" field, but doesn't necessarily succeed.  To have a guarantee of an "id" field, the line:
`        requireNonNull(map.get("id"), "map{id} is null");`
in TraceToken should instead be:
`        requireNonNull(delegate.get("id"), "map{id} is null");`
after the copy into the ImmutableMap is made.

Very much a corner case though...

event/src/main/java/com/proofpoint/event/client/EventJsonSerializer.java
```diff
@@ -18,27 +18,40 @@
...
 
-    public EventJsonSerializer(NodeInfo nodeInfo, @Nullable String token, EventTypeMetadata<T> eventTypeMetadata)
+    EventJsonSerializer(NodeInfo nodeInfo, @Nullable TraceToken token, EventTypeMetadata<T> eventTypeMetadata)
     {
-        this.token = token;
-        this.eventTypeMetadata = checkNotNull(eventTypeMetadata, "eventTypeMetadata is null");
+        if (token == null || eventTypeMetadata.getTraceTokenField() != null) {
+            this.token = null;
+        }
+        else if (token.size() == 1) {
```
ttockpfpt: This feels wrong, as it's breaking TraceToken encapsulation and using knowledge about how it internally stores things.  Better to have methods on TraceToken that know how to marshal/unmarshal into a string, rather than have multiple places outside of TraceToken doing it.

ttockpfpt:
It makes more sense to me know than it did late Sunday night.  :-)  Go ahead and merge it if you want, though I think moving the TraceToken marshal/unmarshal logic into TraceToken would make sense.

event/src/main/java/com/proofpoint/event/client/JsonEventSerializer.java
johngmyers: I'm not particularly worried about that use case. It's not correct usage to change the thread name while a trace token is registered. I don't see this as sufficient justification to expose a public factory for `TraceToken`

event/src/main/java/com/proofpoint/event/client/EventJsonSerializer.java
johngmyers: I really don't want a dependency from trace-token to json, as trace-token is so low-level.

johngmyers:
@ttockpfpt please mark this PR as approved.

johngmyers:
Rebased, squashed fixup commits, and merged

PR #581, opened 2017-02-02
Don't have Platform ExceptionMappers pick up media types from resource method annotations.

johngmyers:
Don't have Platform ExceptionMappers pick up media types from resource method annotations.


johngmyers:
Rebased and merged

PR #587, opened 2017-03-23
Blocking timeout fixes

johngmyers:


CHANGES
```diff
@@ -6,6 +6,8 @@ Platform 1.54
   it was being included in the launcher.log twice. We removed it from its old
   location.
 
+* [Bug] Set the blocking timeout to be more than the idle timeout.
```
gwittel: Jetty docs claim the ```setBlockingTimeout``` value is already in addition to the ```setIdleTimeout``` (which seems misleading based on what I can see).  Is there a jetty bug somewhere we need to track this to resolution (i.e. when to remove the hack)?

gwittel:
LGTM.

CHANGES
johngmyers: It seems to be an ambiguity in language. The blocking timeout feature never looks at the idle timeout. What they apparently meant is that both timeouts apply independently.

johngmyers:
Rebased and merged

PR #591, opened 2017-04-21
Allow Trace Logging its needed for the slf4j adapter to write trace …

JasonQuinn:
Adding a Trace and All logging level.
Trace is needed so the slf4j-to-JUL adapter can be set to write TRACE messages to the log.

### DIFF
--- a/log/src/test/java/com/proofpoint/log/TestLogger.java
+++ b/log/src/test/java/com/proofpoint/log/TestLogger.java
@@ -144,6 +144,14 @@ public class TestLogger
     }
 
     @Test
+    public void testTraceShortCircuit()
+    {
+        inner.setLevel(Level.OFF);
+        logger.trace("hello");
+        assertTrue(handler.isEmpty());
+    }
+
+    @Test
     public void testDebugShortCircuit()
     {
         inner.setLevel(Level.OFF);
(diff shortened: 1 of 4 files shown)


log/src/test/java/com/proofpoint/log/TestLogger.java
```diff
@@ -195,6 +210,18 @@ public void testErrorWithThrowableNoMessageShortCircuit()
```
johngmyers: `testTraceShortCircuit()` ?

log/src/test/java/com/proofpoint/log/TestLogging.java
```diff
@@ -124,6 +124,14 @@ public void testPropagatesLevelsHierarchical()
...
+        logging.setLevel("testPropagatesLevelsHierarchical", Level.TRACE);
+        assertTrue(logger.isDebugEnabled());
+        assertTrue(logger.isInfoEnabled());
+
+        logging.setLevel("testPropagatesLevelsHierarchical", Level.ALL);
+        assertTrue(logger.isDebugEnabled());
+        assertTrue(logger.isInfoEnabled());
     }
 
     @Test
```
johngmyers: Also add to `testPropagatesLevels()`

johngmyers:
Squashed, added CHANGES, and merged.

PR #593, opened 2017-05-19
Implement decorrelated jitter backoff between retries in BalancingHttpClient

johngmyers:


http-client/src/main/java/com/proofpoint/http/client/balancing/BalancingHttpClient.java
```diff
@@ -248,26 +283,40 @@ else if (t instanceof FailureStatusException) {
...
+                                        }
+                                        catch (RuntimeException e1) {
+                                            setException(e1);
+                                        }
+                                    }
                                 }
-                                catch (Exception e2) {
-                                    setException(e2);
+                                catch (Exception e) {
+                                    e.printStackTrace();
```
gwittel: print?

johngmyers: oops

johngmyers: Removed this catch block. Also removed public visibility of new interface/class.

johngmyers:
Squashed the fixups and merged

PR #594, opened 2017-05-24
Implement failure accrual

johngmyers:


CHANGES
```diff
@@ -6,6 +6,10 @@ Platform 1.56
   decorrelated jitter. The minimum and maximum delays between attempts
   are configurable.
 
+  The HttpServiceBalancer now implements failure accrual. Destinations
```
gwittel: Please make the backoff window and failure count configurable.

johngmyers: Making them configurable is problematic.

If there are multiple `BalancingHttpClient`s all using the same underlying discovered service type, then they share the same `HttpServiceBalancer`. So the config could not be hung off of the _name_.http-client property space, but would have to be hung off of the discovery._type_ namespace or some new namespace (as it has little to do with discovery service).

The mechanics of getting the appropriate config object to `HttpServiceBalancerProvider` are also inordinately complex.

It doesn't seem that configurability of these is particularly important; the main purpose of the feature is to keep retry quotas from being so dangerous.

gwittel: That makes sense.  I'm a little worried that at volume this may cause too much flapping, but hard to say.

Another question - If I have to debug a  HTTP client that's throwing failures or backlogging, it would be good to know how/when the new backoff behavior is contributing to the overall system behavior.    Right now we just have failure counts which imply a backoff is occurring.

It would be helpful to have stats covering the state transitions at some level.

johngmyers: Finagle has four counters:

* removals
* revivals
* probes
* removal_for_ms

A behavior difference between this and Finagle is that this code will send traffic to dead/probeClosed instances if there are no alternatives whereas Finagle will fail the request immediately.

These counters seem appropriate, modulo the way they represent the cumulative time.

Finagle doesn't appear to generate any stats on retry backoffs. It does keep a counter of the number of times a retry was suppressed due to quota.

johngmyers:
Implemented metrics and configuration. Disabled by default.

johngmyers:
squashed and merged

PR #595, opened 2017-05-26
Create JaxrsModule.adminOnlyJaxrsModule()

johngmyers:


jaxrs/src/main/java/com/proofpoint/jaxrs/JaxrsModule.java
```diff
@@ -271,48 +158,220 @@ else if (type.isAnnotationPresent(Path.class)) {
...
-            ServiceLocator locator = locatorReference.get();
-            locator.inject(object);
-            locator.postConstruct(object);
-            return object;
+            binder.disableCircularProxies();
+
+            jaxrsBinder(binder).bind(JsonMapper.class);
+            jaxrsBinder(binder).bind(SmileMapper.class);
+            jaxrsBinder(binder).bind(TimingResourceDynamicFeature.class);
+            jaxrsBinder(binder).bind(DisallowOptionsModelProcessor.class);
```
lamielle: If I read properly, this is only relevant to JAX-RS applications, so maybe it doesn't make sense in the common module?

Similar comment for TimingResourceDynamicFeature.

jaxrs/src/main/java/com/proofpoint/jaxrs/JaxrsModule.java
```diff
@@ -271,48 +158,220 @@ else if (type.isAnnotationPresent(Path.class)) {
...
+            jaxrsBinder(binder).bindInjectionProvider(ClientInfo.class).to(ClientInfoSupplier.class);
+            jaxrsBinder(binder).bindAdmin(ParsingExceptionMapper.class);
+            jaxrsBinder(binder).bindAdmin(QueryParamExceptionMapper.class);
+            jaxrsBinder(binder).bindAdmin(OverrideMethodFilter.class);
+            jaxrsBinder(binder).bindAdmin(ThreadDumpResource.class);
+
+            newSetBinder(binder, Object.class, JaxrsResource.class).permitDuplicates();
```
lamielle: I don't think these three lines are needed for a non-JAX-RS application either.  At least, the module I implemented that does effectively what you're doing here did not need them.

jaxrs/src/main/java/com/proofpoint/jaxrs/JaxrsModule.java
johngmyers: I thought it disallowed OPTIONS on the admin port as well, but apparently not.

lamielle:
I pulled your branch and after a brief test it seems using adminOnlyJaxrsModule is working as expected!

jaxrs/src/main/java/com/proofpoint/jaxrs/JaxrsModule.java
johngmyers: Need these to create the Application. I believe non-`Resource` singletons such as the `InjectionProvider`s also apply to the admin port.

johngmyers:
Squashed and merged

PR #596, opened 2017-05-30
Create static BalancingHttpClient binder to parameterized annotation

johngmyers:


http-client/src/main/java/com/proofpoint/http/client/HttpClientBinder.java
```diff
@@ -108,6 +105,18 @@ public BalancingHttpClientBindingBuilder bindBalancingHttpClient(String name, Cl
         return new BalancingHttpClientBindingBuilder(binder, annotation, delegateBindingBuilder);
     }
 
+    public BalancingHttpClientBindingBuilder bindBalancingHttpClient(String name, Annotation annotation, String serviceName, Set<URI> baseUris)
```
lamielle: This is roughly the signature I was looking for.  The only detail that's not clear to me is the difference/purpose of the 'name' parameter vs. 'serviceName'.  Could you define those and how they're different?

johngmyers: name is the name of the HttpClient, used to prefix its configuration properties.

'serviceName' is the name of the service, used in the ServiceClient metrics that are reported. For a discovered service, this would be the discovery "type". This would usually be the same as the annotation value, but the code can't assume the annotation has a "value" method.

johngmyers:
Rebased and merged

PR #598, opened 2017-06-08
Create StaticDiscoveryModule

johngmyers:


configuration/pom.xml
```diff
@@ -43,6 +43,12 @@
             <version>3.1</version>
         </dependency>
 
+        <dependency>
```
alinav: do we have to bind google auto-value?
Immutables.org might be better? http://marxsoftware.blogspot.com/2016/06/lombok-autovalue-immutables.html

johngmyers: Immutables would have made excluding the `Key` attribute from `equals()` slightly easier, but AutoValue is used extensively elsewhere in Platform.

CHANGES
```diff
@@ -9,6 +9,12 @@ Platform 1.57
   "service-balancer.SERVICE.consecutive-failures" configuration option. A
   value of "5" is recommended.
 
+* Discovery
+
+  We created a StaticDiscoveryModule which does not talk to the discovery
+  service. It instead obtains service location from configuration, using
+  the "service-balancer.SERVICETYPE.uri" config option.
```
gwittel: Please include a note that it supports comma separated list of URIs

johngmyers:
Tweaked CHANGES and merged

PR #614, opened 2017-09-25
Upgrade libs

johngmyers:


johngmyers:
Oops, we previously had an issue with bval-jsr 1.1.1

PR #616, opened 2017-10-04
Allow GCs other than ConcMarkSweep in jvm.config

johngmyers:


launcher/src/main/java/com/proofpoint/launcher/Main.java
```diff
@@ -328,6 +313,33 @@ else if (line.matches("(?i)\\s*#\\s*allow\\s+spaces\\s*")) {
...
+            javaArgs.add("-server");
+            if (!gcSpecified) {
+                javaArgs.add("-XX:+UseConcMarkSweepGC");
+                javaArgs.add("-XX:+ExplicitGCInvokesConcurrent");
+            }
+            javaArgs.add("-XX:+HeapDumpOnOutOfMemoryError");
+            javaArgs.add("-XX:HeapDumpPath=var");
+            javaArgs.add("-XX:+AggressiveOpts");
+            javaArgs.add("-XX:+DoEscapeAnalysis");
+            javaArgs.add("-XX:+UseCompressedOops");
```
cpuchakayaka: Is there a need to have this jvm option, as this is default now in most JVM's, unless there is need to support older JVM's

johngmyers: I leave those in primarily to make it easier to show that the legacy entries in jvm.config files aren't needed anymore.

PR #618, opened 2017-10-09
Add support for building Docker images

johngmyers:


library/pom.xml
```diff
@@ -1365,5 +1375,127 @@
...
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-enforcer-plugin</artifactId>
+                        <configuration>
+                            <rules combine.children="append">
+                                <requireProperty>
+                                    <property>main-class</property>
+                                    <message>The main-class property must be specified when building a platform tarball.</message>
```
gwittel: tarball?

johngmyers: Fixed

gwittel:
It looks like the baseline fabric8 docker config is to push to the public docker.io registry by default.  (unless there's some magic around the `dev-docker-local/` prefix).

While I wouldn't want to hard code a registry name, it seems like this default will make it easy for people to accidentally push private images out to docker.io. Should we require making the registry explicit by one of the fabric8 supported methods (property, environment)?  Or a build profile (public v private docker image)?

johngmyers:
The build machines use an environment variable to point to our internal registry. We don't want to encode the registry in the POM. It doesn't appear possible to specify the project prefix in an environment variable, so that has to go in the POM.

I expect deploy/push operations to only happen on build machines. I think the risk of a successful accidental push to docker.io is pretty low.

gwittel:
Given an accidental push requires one to run `mvn deploy`, have Maven repo privs, docker push privs, it seems OK.  If it helps, maybe a readme blurb on the default behavior/configuring the docker push target (or a reference to the fabric8 docs).

johngmyers:
Squashed fixup commit and merged

PR #619, opened 2017-10-10
Query Kubernetes service discovery for discovery servers

johngmyers:


CHANGES
```diff
@@ -14,6 +14,11 @@ Platform 1.62
   The image installs the service under the "/service" directory and runs the
   service as the user "service".
 
+* Discovery
+
+  If neither "service-inventory.uri" or "testing.discovery.uri" are configured
+  then the discovery client will query Kubernetes for the discovery service.
```
gwittel: We should clarify that the query is made by DNS lookup on the short host `discovery`

PR #624, opened 2017-10-19
Miscellaneous Java 9 compatibility fixes

johngmyers:


jmx/src/test/java/com/proofpoint/jmx/TestJmxAgent.java
```diff
@@ -17,15 +17,15 @@
 public class TestJmxAgent
```
gwittel: Any reason to not rename this class to match the `JmxAgent8` naming?

johngmyers: It might make it slightly easier to track future Airlift changes.

johngmyers:
Rebased and merged

PR #626, opened 2017-11-29
add support for multiple config files

rajatjindal:
Initial draft for multiple config files. Please let me know if this sounds like an OK approach.

I am still testing it more to make sure I don't break anything and its backward compatible.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,9 @@
+Platform 1.65
+
+* Configuration
+
+  - Supports reading configs from multiple files. Multiple files can be provided using comma separated names
+
 Platform 1.64
 
 * Docker
--- a/CHANGES
+++ b/CHANGES
@@ -2,7 +2,7 @@ Platform 1.65
 
 * Configuration
 
-  - Supports reading configs from multiple files. Multiple files can be provided using comma separated names
+  - Add an optional --secret-config flag. If provided, read the secret config properties from it. In a future version, it will be mandatory to use @ConfigSecuritySensitive annotation on properties defined in this file
 
 Platform 1.64
 
--- a/CHANGES
+++ b/CHANGES
@@ -2,7 +2,8 @@ Platform 1.65
 
 * Configuration
 
-  - Add an optional --secret-config flag. If provided, read the secret config properties from it. In a future version, it will be mandatory to use @ConfigSecuritySensitive annotation on properties defined in this file
+  Add an optional --secret-config flag. If provided, read the secret config properties from it.  In a 
+  future version, it will be mandatory to use @ConfigSecuritySensitive annotation on configs used from this file
 
 Platform 1.64
 
--- a/bootstrap/src/main/java/com/proofpoint/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/com/proofpoint/bootstrap/Bootstrap.java
@@ -44,8 +44,17 @@ import com.proofpoint.node.ApplicationNameModule;
 import com.proofpoint.node.NodeInfo;
 
 import java.io.PrintWriter;
-import java.util.*;
+import java.util.Arrays;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
+import java.util.TreeMap;
+import java.util.TreeSet;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import static com.google.common.base.Preconditions.checkNotNull;
@@ -322,11 +331,11 @@ public class Bootstrap
         if (requiredConfigurationProperties == null) {
             log.info("Loading configuration");
             String config = System.getProperty("config");
-            if (config == null) {
-                config = "";
+            Collection<String> files = new ArrayList<>();
+            if (config != null) {
+                files = Arrays.asList(config.split(","));
             }
 
-            List<String> files = Arrays.asList(config.split(","));
             builder = builder
                     .withFiles(files)
                     .withSystemProperties();
--- a/bootstrap/src/main/java/com/proofpoint/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/com/proofpoint/bootstrap/Bootstrap.java
@@ -44,14 +44,8 @@ import com.proofpoint.node.ApplicationNameModule;
 import com.proofpoint.node.NodeInfo;
 
 import java.io.PrintWriter;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.Map.Entry;
-import java.util.Set;
-import java.util.TreeMap;
-import java.util.TreeSet;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import static com.google.common.base.Preconditions.checkNotNull;
@@ -327,8 +321,14 @@ public class Bootstrap
         }
         if (requiredConfigurationProperties == null) {
             log.info("Loading configuration");
+            String config = System.getProperty("config");
+            if (config == null) {
+                config = "";
+            }
+
+            List<String> files = Arrays.asList(config.split(","));
             builder = builder
-                    .withFiles(System.getProperty("config"))
+                    .withFiles(files)
                     .withSystemProperties();
         }
         else {
--- a/bootstrap/src/main/java/com/proofpoint/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/com/proofpoint/bootstrap/Bootstrap.java
@@ -327,12 +327,10 @@ public class Bootstrap
         }
         if (requiredConfigurationProperties == null) {
             log.info("Loading configuration");
-            String config = System.getProperty("config");
-            String secretConfig = System.getProperty("secret-config");
 
             builder = builder
-                    .withFile(config)
-                    .withFile(secretConfig)
+                    .withFile(System.getProperty("config"))
+                    .withFile(System.getProperty("secret-config"))
                     .withSystemProperties();
         }
         else {
--- a/bootstrap/src/main/java/com/proofpoint/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/com/proofpoint/bootstrap/Bootstrap.java
@@ -328,7 +328,7 @@ public class Bootstrap
         if (requiredConfigurationProperties == null) {
             log.info("Loading configuration");
             builder = builder
-                    .withFile(System.getProperty("config"))
+                    .withFiles(System.getProperty("config"))
                     .withSystemProperties();
         }
         else {
--- a/bootstrap/src/main/java/com/proofpoint/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/com/proofpoint/bootstrap/Bootstrap.java
@@ -327,7 +327,6 @@ public class Bootstrap
         }
         if (requiredConfigurationProperties == null) {
             log.info("Loading configuration");
-
             builder = builder
                     .withFile(System.getProperty("config"))
                     .withFile(System.getProperty("secret-config"))
--- a/bootstrap/src/main/java/com/proofpoint/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/com/proofpoint/bootstrap/Bootstrap.java
@@ -331,13 +331,11 @@ public class Bootstrap
         if (requiredConfigurationProperties == null) {
             log.info("Loading configuration");
             String config = System.getProperty("config");
-            Collection<String> files = new ArrayList<>();
-            if (config != null) {
-                files = Arrays.asList(config.split(","));
-            }
+            String secretConfig = System.getProperty("secret-config");
 
             builder = builder
-                    .withFiles(files)
+                    .withFile(config)
+                    .withFile(secretConfig)
                     .withSystemProperties();
         }
         else {
--- a/bootstrap/src/main/java/com/proofpoint/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/com/proofpoint/bootstrap/Bootstrap.java
@@ -44,11 +44,8 @@ import com.proofpoint.node.ApplicationNameModule;
 import com.proofpoint.node.NodeInfo;
 
 import java.io.PrintWriter;
-import java.util.Arrays;
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.HashMap;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
--- a/bootstrap/src/test/java/com/proofpoint/bootstrap/TestBootstrap.java
+++ b/bootstrap/src/test/java/com/proofpoint/bootstrap/TestBootstrap.java
@@ -60,6 +60,7 @@ public class TestBootstrap
     public void setup()
     {
         System.clearProperty("config");
+        System.clearProperty("secret-config");
         System.clearProperty("property");
     }
 
--- a/bootstrap/src/test/java/com/proofpoint/bootstrap/TestBootstrap.java
+++ b/bootstrap/src/test/java/com/proofpoint/bootstrap/TestBootstrap.java
@@ -126,6 +126,84 @@ public class TestBootstrap
     }
 
     @Test
+    public void testMultipleConfigFiles()
+            throws Exception
+    {
+        String propertyFiles = new StringBuilder()
+                .append(Resources.getResource("simple-config.properties").getFile())
+                .append(",")
+                .append(Resources.getResource("second-config.properties").getFile())
+                .toString();
+
+        System.setProperty("config", propertyFiles);
+        Injector injector = bootstrapApplication("test-application")
+                .doNotInitializeLogging()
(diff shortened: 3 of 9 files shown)


rajatjindal:
Testcases:

1. No config file provided
Error: Config file is missing
Tested: true

2. Config file provided, but not present
Error: Config file is missing
Tested: true

3. Config file provided, present, but no values
Error: Mandatory parameter node.environment missing
Tested: true

4. Config file provided, present with mandatory values
Result: starts successfully
Tested: true

5. Multiple config files, present, with no duplicate values
starts successfuly
Tested: true

6. Multiple config files, present, duplicate values
Error: property 'store.ttl' already defined in file <filename>
Tested: true

7. Multiple config files, present, duplicate filenames
Error: Duplicate config file
Tested: true

8. Multiple config files, some not present
Error: Config file is missing
Tested: true

rajatjindal:
9. /path/to/valid-file,,,/path/to/valid-file2
Result: read configs from  /path/to/valid-file and  /path/to/valid-file2
Tested: true

rajatjindal:
@johngmyers can you please review these changes

configuration/src/main/java/com/proofpoint/configuration/ConfigurationFactoryBuilder.java
```diff
@@ -75,8 +71,44 @@ public synchronized Object put(Object key, Object value) {
...
+    }
+
+    /**
+     * Loads properties from list of comma separated files
+     *
+     * @param files comma separated list of config files to be loaded
+     * @return self
+     * @throws java.io.IOException errors
+     */
+    public ConfigurationFactoryBuilder withFiles(@Nullable final String files)
```
johngmyers: Should take a `Collection<String>` at this layer, not a `String` with commas in it.

configuration/src/main/java/com/proofpoint/configuration/ConfigurationFactoryBuilder.java
```diff
@@ -52,13 +52,9 @@
      * @return self
      * @throws java.io.IOException errors
      */
-    public ConfigurationFactoryBuilder withFile(@Nullable final String path)
```
johngmyers: Can't remove public APIs without deprecation period. We should probably leave this in anyway.

configuration/src/main/java/com/proofpoint/configuration/ConfigurationFactoryBuilder.java
```diff
@@ -75,8 +71,44 @@ public synchronized Object put(Object key, Object value) {
...
+            if (!fileNames.add(file)) {
+                errors.add(format("Duplicate config file '%s'", file));
+                continue;
+            }
+
+            Properties properties = this.fromFile(file);
+            for (Map.Entry<Object, Object> entry : properties.entrySet()) {
+                if (propertyMap.containsKey(entry.getKey().toString())) {
+                    errors.add(format("property '%s' already defined in file '%s'",
+                            entry.getKey().toString(), propertyMap.get(entry.getKey().toString())));
```
johngmyers: I think the previous approach of logging the name of the later file is sufficient.

configuration/src/main/java/com/proofpoint/configuration/ConfigurationFactoryBuilder.java
```diff
@@ -75,8 +71,44 @@ public synchronized Object put(Object key, Object value) {
...
+    {
+        if (files == null) {
+            return this;
+        }
+
+        Set<String> fileNames = new HashSet<>();
+        Map<String, String> propertyMap = new HashMap<>();
+
+        for(String file: files.split(",")) {
+            if (!fileNames.add(file)) {
```
johngmyers: I'm dubious about the value of this check. If the file has any properties, the duplicate-properties check will catch this. If a refactor removed the check, I would be okay with that.

launcher/src/main/java/com/proofpoint/launcher/Main.java
```diff
@@ -173,8 +173,16 @@ public final void run()
...
                 launcherArgs.add("--config");
-                launcherArgs.add(new File(configPath).getAbsolutePath());
+                StringBuilder builder = new StringBuilder();
+                for (String file:configPath.split(",")) {
+                    builder.append(new File(file).getAbsolutePath());
+                    builder.append(",");
+                }
+
+                configPath = builder.toString().replace(",$", "");
+                launcherArgs.add(configPath);
```
johngmyers: Something like:

```java
Arrays.stream(configPath.split(","))
                        .map(path -> new File(path).getAbsolutePath())
                        .collect(Collectors.joining(","))
```

launcher/src/main/java/com/proofpoint/launcher/Main.java
```diff
@@ -173,8 +173,16 @@ public final void run()
                 configPath = installPath + "/etc/config.properties";
             }
             else {
+                //put validation here
```
johngmyers: What validation?

launcher/src/main/java/com/proofpoint/launcher/Main.java
```diff
@@ -287,11 +295,14 @@ protected void start(List<String> args, boolean daemon)
                 System.exit(0);
             }
 
-            if (!new File(configPath).exists()) {
-                System.err.println("Config file is missing: " + configPath);
-                System.exit(STATUS_CONFIG_MISSING);
+            for (String file:configPath.split(",")) {
```
johngmyers: style: spaces around :

launcher/src/main/java/com/proofpoint/launcher/Main.java
```diff
@@ -287,11 +295,14 @@ protected void start(List<String> args, boolean daemon)
...
+                if (!new File(file).exists()) {
+                    System.err.println("Config file is missing: " + file);
+                    System.exit(STATUS_CONFIG_MISSING);
+                }
             }
 
+
```
johngmyers: nit: erroneous blank line

configuration/src/main/java/com/proofpoint/configuration/ConfigurationFactoryBuilder.java
rajatjindal: it was just to reduce the noise in logging, if we have a file with 10 properties, and provided same file 2 times, the current check will print 10 times about the duplicate file

configuration/src/main/java/com/proofpoint/configuration/ConfigurationFactoryBuilder.java
rajatjindal: we need to create one properties instance for storing properties from all files, which means we need to initialize the Properties object even before we start calling the iterator on files.

and because we are overriding the function put from Properties class, the variable 'path' is not available for logging there.

configuration/src/main/java/com/proofpoint/configuration/ConfigurationFactoryBuilder.java
rajatjindal: I'll make this change, and leave this there

configuration/src/main/java/com/proofpoint/configuration/ConfigurationFactoryBuilder.java
johngmyers: You could wrap the combined Properties with a fowarding Properties that does the duplicate check and captures path. Or you could subclass Properties with something that can have the current path pushed into it.

bootstrap/src/main/java/com/proofpoint/bootstrap/Bootstrap.java
```diff
@@ -44,14 +44,8 @@
 import com.proofpoint.node.NodeInfo;
 
 import java.io.PrintWriter;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
johngmyers: Not permitted per style. Please load the Proofpoint style into your IDE.

bootstrap/src/main/java/com/proofpoint/bootstrap/Bootstrap.java
```diff
@@ -327,8 +321,14 @@ public Injector initialize()
         }
         if (requiredConfigurationProperties == null) {
             log.info("Loading configuration");
+            String config = System.getProperty("config");
+            if (config == null) {
+                config = "";
```
johngmyers: This would result in a single file named "". It should result in zero files.

configuration/src/main/java/com/proofpoint/configuration/ConfigurationFactoryBuilder.java
```diff
@@ -80,6 +81,44 @@ public synchronized Object put(Object key, Object value) {
...
+     * Loads properties from list of comma separated files
+     *
+     * @param files comma separated list of config files to be loaded
+     * @return self
+     * @throws java.io.IOException errors
+     */
+    public ConfigurationFactoryBuilder withFiles(final Collection<String> files)
+            throws IOException
+    {
+        if (files.isEmpty()) {
```
johngmyers: Unnecessary

configuration/src/main/java/com/proofpoint/configuration/ConfigurationFactoryBuilder.java
```diff
@@ -80,6 +81,44 @@ public synchronized Object put(Object key, Object value) {
...
     }
 
+    /**
+     * Loads properties from list of comma separated files
+     *
+     * @param files comma separated list of config files to be loaded
+     * @return self
+     * @throws java.io.IOException errors
+     */
+    public ConfigurationFactoryBuilder withFiles(final Collection<String> files)
```
johngmyers: nit: final not necessary

configuration/src/main/java/com/proofpoint/configuration/ConfigurationFactoryBuilder.java
```diff
@@ -80,6 +81,44 @@ public synchronized Object put(Object key, Object value) {
...
+            throws IOException
+    {
+        if (files.isEmpty()) {
+            return this;
+        }
+
+        Set<String> fileNames = new HashSet<>();
+        Map<String, String> propertyMap = new HashMap<>();
+
+        for(String file: files) {
```
johngmyers: Spacing style is wrong.

Shouldn't this just delegate to `withFile()` ?

launcher/src/main/java/com/proofpoint/launcher/Main.java
```diff
@@ -174,7 +174,19 @@ public final void run()
...
+                    if (file == "") {
+                        continue;
+                    }
+
+                    builder.append(new File(file).getAbsolutePath());
+                    builder.append(",");
+                }
+
+                configPath = builder.toString().replace("(.*),$", "$1");
+                launcherArgs.add(configPath);
```
johngmyers: Comment not addressed and GitHub suppressing it for some reason:

Something like:

```java
Arrays.stream(configPath.split(","))
                        .map(path -> new File(path).getAbsolutePath())
                        .collect(Collectors.joining(","))
```

bootstrap/src/main/java/com/proofpoint/bootstrap/Bootstrap.java
rajatjindal: fixed. thnx

bootstrap/src/main/java/com/proofpoint/bootstrap/Bootstrap.java
rajatjindal: fixed, thnx

configuration/src/main/java/com/proofpoint/configuration/ConfigurationFactoryBuilder.java
rajatjindal: fixed

configuration/src/main/java/com/proofpoint/configuration/ConfigurationFactoryBuilder.java
rajatjindal: fixed

launcher/src/main/java/com/proofpoint/launcher/Main.java
rajatjindal: wow, this is so much cleaner. thanks John

launcher/src/main/java/com/proofpoint/launcher/Main.java
rajatjindal: it was a typo there

configuration/src/main/java/com/proofpoint/configuration/ConfigurationFactoryBuilder.java
rajatjindal: done. not sure how all these comments got suppressed

CHANGES
```diff
@@ -1,3 +1,9 @@
+Platform 1.65
+
+* Configuration
+
+  - Supports reading configs from multiple files. Multiple files can be provided using comma separated names
```
johngmyers: Need to fix this as well

bootstrap/src/main/java/com/proofpoint/bootstrap/Bootstrap.java
```diff
@@ -327,8 +330,12 @@ public Injector initialize()
         }
         if (requiredConfigurationProperties == null) {
             log.info("Loading configuration");
+            String config = System.getProperty("config");
+            String secretConfig = System.getProperty("secret-config");
+
             builder = builder
-                    .withFile(System.getProperty("config"))
+                    .withFile(config)
+                    .withFile(secretConfig)
```
johngmyers: No reason not to inline the variables. Also optimize the imports.

bootstrap/src/test/java/com/proofpoint/bootstrap/TestBootstrap.java
```diff
@@ -125,6 +125,24 @@ public void testConfigFile()
         assertEquals(simpleConfig.getProperty(), "value");
     }
 
+    @Test
+    public void testSecretConfig()
+            throws Exception
+    {
+        System.setProperty("config", Resources.getResource("simple-config.properties").getFile());
+        System.setProperty("secret-config", Resources.getResource("secret-config.properties").getFile());
```
johngmyers: Clear the new property in `setup()` (or a newly-created `@AfterMethod(alwaysRun = true)`)

launcher/src/main/java/com/proofpoint/launcher/Main.java
```diff
@@ -174,7 +179,11 @@ public final void run()
             }
             else {
                 launcherArgs.add("--config");
-                launcherArgs.add(new File(configPath).getAbsolutePath());
+                launcherArgs.add(configPath);
```
johngmyers: keep the `getAbsolutePath()`

launcher/src/main/java/com/proofpoint/launcher/Main.java
```diff
@@ -174,7 +179,11 @@ public final void run()
...
+            }
+            if (secretConfigPath != null) {
+                launcherArgs.add("--secret-config");
+                launcherArgs.add(secretConfigPath);
```
johngmyers: Do `getAbsolutePath()` here too. This keeps things working even after a change-directory.

launcher/src/main/java/com/proofpoint/launcher/Main.java
```diff
@@ -287,11 +301,17 @@ protected void start(List<String> args, boolean daemon)
                 System.exit(0);
             }
 
+
```
johngmyers: nit: spurious

CHANGES
```diff
@@ -1,3 +1,9 @@
+Platform 1.65
+
+* Configuration
+
+  - Add an optional --secret-config flag. If provided, read the secret config properties from it. In a future version, it will be mandatory to use @ConfigSecuritySensitive annotation on properties defined in this file
```
johngmyers: Please wrap to no longer than 80 columns

launcher/src/main/java/com/proofpoint/launcher/Main.java
rajatjindal: yep, that was unintentional

bootstrap/src/main/java/com/proofpoint/bootstrap/Bootstrap.java
rajatjindal: done

bootstrap/src/test/java/com/proofpoint/bootstrap/TestBootstrap.java
rajatjindal: clearing in setup() function done

launcher/src/main/java/com/proofpoint/launcher/Main.java
```diff
@@ -58,6 +58,7 @@
     private static final int STATUS_INVALID_ARGS = 2;
     private static final int STATUS_UNSUPPORTED = 3;
     private static final int STATUS_CONFIG_MISSING = 6;
+    private static final int STATUS_SECRET_CONFIG_MISSING = 7;
```
johngmyers: These status codes come from the initscript contract: https://refspecs.linuxfoundation.org/LSB_3.0.0/LSB-PDA/LSB-PDA/iniscrptact.html

So use STATUS_CONFIG_MISSING instead of code 7, which means something else.

rajatjindal:
Not sure why this does not show as merged.

johngmyers:
Squashed and merged

PR #629, opened 2017-12-03
fix bootstrap/test to use secrets-config instead of secret-config

rajatjindal:
@johngmyers 

This slipped through the review I did. my bad.


rajatjindal:
thanks @johngmyers 

when you get chance, can you please merge this change, and release new version of platform

PR #634, opened 2018-01-29
Add support for jvm.properties

johngmyers:


CHANGES
```diff
@@ -1,5 +1,13 @@
 Platform 1.67
 
+* Launcher
+
+  We added support for an etc/jvm.properties file as an alternative for the
+  jvm.config file. Each property in that Java properties file is used to
+  construct a Java command-line argument by concatenating the property
+  key and value (without the = in between). The "-classpath" key is handled
+  as a special case, resulting in the key and value being separate arguments.
+
```
gwittel: Should we make it explicit that support is for either, not both?  Or go to the level of throwing a warning/error if both are supplied?

johngmyers: It ignores jvm.config if jvm.properties is present. It does state that it is an alternative. Do you have a suggestion for improved wording?

gwittel: I think the `as an alternative` wording is vague as far as what really happens -- jvm.config is ignored when jvm.properties is present.

Perhaps just adding a clarifying 2nd sentence or clause along the lines of:

> We added support for an etc/jvm.properties file as an alternative to the jvm.config file.  When jvm.properties is present, it will be used instead of jvm.config. .....

johngmyers: Added fixup commit.

johngmyers:
Rebased and merged

PR #638, opened 2018-02-19
adding before_install step

vsilverman:


johngmyers:
There doesn't seem to be anything here of value. The extra step with library is unnecessary and I don't see the point of the before_install step.

johngmyers:
Are you perhaps trying to fish out credentials from Travis?

vsilverman:
Proposing to add README and providing it's initial content to help building the platform from the sources.

johngmyers:
Nothing here of value. Declining.

PR #642, opened 2018-03-08
Read any etc/secrets.properties file by default

johngmyers:


launcher/src/main/java/com/proofpoint/launcher/Main.java
```diff
@@ -251,6 +251,12 @@ public final void run()
                 launcherArgs.add("--config");
                 launcherArgs.add(new File(configPath).getAbsolutePath());
             }
+            if (secretsConfigPath == null) {
+                secretsConfigPath = installPath + "/etc/secrets.properties";
+                if (!new File(secretsConfigPath).exists()) {
+                    secretsConfigPath = null;
+                }
+            }
             if (secretsConfigPath != null) {
```
gwittel: A late question -- Now or in the future, do we want to do any permissions enforcement like refuse to read unless its 0600 (or some strict equivalent)?  This would require coordination with the deploy code.

johngmyers: Don't think so. This is primarily intended for use within containers, so not an issue there.

PR #644, opened 2018-03-22
Make parsing of boolean configuration properties stricter.

johngmyers:


configuration/src/main/java/com/proofpoint/configuration/ConfigurationFactory.java
```diff
@@ -554,7 +554,13 @@ private static Object coerce(Class<?> type, String value)
...
             }
             else if (Boolean.class.isAssignableFrom(type) || Boolean.TYPE.isAssignableFrom(type)) {
-                return Boolean.valueOf(value);
+                if ("true".equalsIgnoreCase(value)) {
+                    return true;
+                }
+                if ("false".equalsIgnoreCase(value)) {
+                    return false;
+                }
+                return null;
```
sorabhlall: Just curious, will it not make more sense to default it to false?

johngmyers: That's what it currently does. That makes values like "ture" be interpreted as false.

sorabhlall: Thanks, got it. I was thinking more about potential NPE.

PR #651, opened 2018-05-11
Extend forwarded for

johngmyers:


http-server/src/main/java/com/proofpoint/http/server/CidrSet.java
```diff
@@ -0,0 +1,98 @@
...
+
+import java.net.Inet4Address;
+import java.net.InetAddress;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class CidrSet
```
gwittel: Maybe Cidr4Set to match the v4 content?

johngmyers: The interface should be capable of handling IPv6 as well. Notice that `containsAddress` takes an `InetAddress`. It's just that the implementation doesn't currently support IPv6.

CHANGES
```diff
@@ -1,5 +1,9 @@
 Platform 1.74
 
+* We added the "http-server.internal-networks" configuration property.
+  This takes a comma-separated list of IPv4 CIDR blocks which are additionally
+  trusted when processing the X-Forwarded-For: header on incoming requests.
```
gwittel: Maybe clarify that this is in addition to the standard private networks and not overriding?

http-server/src/main/java/com/proofpoint/http/server/CidrSet.java
gwittel: Yeah I saw that.  I'm fine just mentioning the caveat and sorting out v6 handling later.

CHANGES
johngmyers: It does say "additionally". How would you word this?

gwittel: Maybe:  "are trusted in addition to private IP address ranges when ..."

johngmyers:
fixup commits pushed

johngmyers:
Squashed fixup commits and merged

PR #653, opened 2018-05-30
Monitor the logging levels file for changes.  When changed reprocess it.

jon-strayer:


btbytes:
The 500 ms check interval is too small. I would change it to 5 minutes or higher.

log/src/main/java/com/proofpoint/log/LogLevelsFileMonitor.java
```diff
@@ -0,0 +1,61 @@
...
+    private static final Logger logger = Logger.get(LogLevelsFileMonitor.class);
+
+    private final Logging logging;
+    private final Timer timer = new Timer("Scan thread", true);
+    private final LoggingConfiguration config;
+
+    protected LogLevelsFileMonitor(LoggingConfiguration logConfig) {
+        logging = Logging.initialize();
+        config = logConfig;
+        timer.scheduleAtFixedRate(new ScanTask(), config.getScanPeriodInMiliseconds(), config.getScanPeriodInMiliseconds() );
```
lamielle: Extra space before the final parenthesis.  A number of other formatting problems are present too: extra newlines, missing braces around conditional blocks, etc.

log/src/main/java/com/proofpoint/log/Logging.java
```diff
@@ -359,6 +360,11 @@ public void configure(LoggingConfiguration config)
         if (config.getBootstrapLogPath() != null) {
             setupBootstrapLog(config.getBootstrapLogPath());
         }
+
+        if (config.isScanEnabled()) {
+            if (monitor == null)
```
lamielle: Could simplify this to:

    if (config.isScanEnabled() && monitor == null)

Also, conditional blocks without braces should generally be avoided.

log/src/main/java/com/proofpoint/log/LoggingConfiguration.java
```diff
@@ -127,5 +127,24 @@ public LoggingConfiguration setMaxTotalSize(DataSize maxTotalSize)
         this.maxTotalSize = maxTotalSize;
         return this;
     }
+
+    public boolean isScanEnabled() { return scanEnabled;}
+
+    @Config("log.scan-enabled")
+    @ConfigDescription("Set to true to eanble run time log level setting. The default is false.")
```
lamielle: Typo: s/eanble/enable

Also, suggestion for updated description:

    Set to true to enable periodic checking of log level settings.  The default is false.

log/src/main/java/com/proofpoint/log/LoggingConfiguration.java
```diff
@@ -127,5 +127,24 @@ public LoggingConfiguration setMaxTotalSize(DataSize maxTotalSize)
...
+    public LoggingConfiguration setScanEnabled(boolean enabled)
+    {
+        this.scanEnabled = enabled;
+        return this;
+    }
+
+    public long getScanPeriodInMiliseconds() {return scanPeriodInMiliseconds;}
+
+    @Config("log.scan-period-in-miliseconds")
+    @ConfigDescription("Number of miliseconds to wait between scans of the levels file. The default is 5,000.")
```
lamielle: Whatever we settle on for the default, make sure to update this.

log/src/main/java/com/proofpoint/log/LoggingConfiguration.java
```diff
@@ -127,5 +127,24 @@ public LoggingConfiguration setMaxTotalSize(DataSize maxTotalSize)
...
+
+    @Config("log.scan-enabled")
+    @ConfigDescription("Set to true to eanble run time log level setting. The default is false.")
+    public LoggingConfiguration setScanEnabled(boolean enabled)
+    {
+        this.scanEnabled = enabled;
+        return this;
+    }
+
+    public long getScanPeriodInMiliseconds() {return scanPeriodInMiliseconds;}
```
lamielle: Even single line methods should be consistently formatted, so for this method and the one above, add appropriate whitespace so braces are on separate lines.

log/src/main/java/com/proofpoint/log/LoggingConfiguration.java
```diff
@@ -32,6 +30,8 @@
     private int maxHistory = 30;
     private String levelsFile = null;
     private DataSize maxTotalSize = new DataSize(1, Unit.GIGABYTE);
+    private boolean scanEnabled;
+    private long scanPeriodInMiliseconds = 5000;
```
gwittel: Use Duration and not long.

log/src/main/java/com/proofpoint/log/LoggingConfiguration.java
```diff
@@ -32,6 +30,8 @@
     private int maxHistory = 30;
     private String levelsFile = null;
     private DataSize maxTotalSize = new DataSize(1, Unit.GIGABYTE);
+    private boolean scanEnabled;
```
gwittel: Scan seems like an unclear name.  Maybe refresh, or something along those lines?  Or to be more consistent with the other settings, `refresh-levels-file`?

Then we can have `log.refresh-levels-file.enabled` and `log.refresh-levels-file.interval`?  Or to avoid dual settings, a duration of null (not set) means disabled.  Though, in that case we'd have to default to off (null).  I'm not sure how to explicitly set it off in the config (`Duration.valueOf` will reject null and empty string)

log/src/main/java/com/proofpoint/log/LogLevelsFileMonitor.java
```diff
@@ -0,0 +1,61 @@
...
+
+    }
+
+
+    private class ScanTask extends TimerTask {
```
gwittel: We should probably be using ScheduledExecutorService instead.  It won't fire off a ton of checks in the event of delays.

Additionally, future executions will never run if this throws an uncaught exception (e.g. setLevels fails); same issue for ScheduledExecutorService also applies.

PR #659, opened 2018-06-18
Changed the default value of the http-server.accept-queue-size configuration option to 128 from 8000

nr17:
Changed the default value of the "http-server.accept-queue-size" configuration
option to 128 from 8000. The OS can truncate the accept queue size to the platform defaults.
E.g. for linux, the effective server listen backlog queue size cannot be set greater
than /proc/sys/net/core/somaxconn on the system.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,5 +1,14 @@
 Platform 1.77
 
+* HttpServer
+
+  We changed the default value of the "http-server.accept-queue-size"
+  configuration option to 128 from 8000. The OS can truncate the accept queue
+  size to the platform defaults. E.g. for linux, the effective server listen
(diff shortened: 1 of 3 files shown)


CHANGES
```diff
@@ -3441,4 +3450,4 @@ annotations. Here's an example on how to use assertFailsValidation():
 The following issues have been fixed:
 	- Discovery client fails when consuming static announcements
 	- MySQL connection timeouts using the wrong unit (ms vs s)
-	- Jar manifest using the wrong file names when building snapshot versions
+	- Jar manifest using the wrong file names when building snapshot versions
```
johngmyers: Looks like you accidentally removed the last newline in the file.

PR #662, opened 2018-06-27
Add responseCodeFamily tag to metrics with responseCode tag

johngmyers:


CHANGES
```diff
@@ -1,3 +1,9 @@
+Platform 1.81
+
+* We added a "responseCodeFamily" tag containing the first digit of the
```
nr17: Who is the consumer of the responseCodeFamily? Can they not derive it from the response code (divide by 100)?

johngmyers: It is to aid in generating Grafana charts that focus only on failure status codes, for example. Grafana cannot divide by 100.

Issue #670, opened 2018-07-24
Support local configuration for development


gdasbach:
A common issue I've run into is working on development of a platform application and requiring a set of local configuration overrides for development.  Ideally I don't want to have to edit the config.properties file to hold my local overrides (and accidentally check something in or deal with merge conflicts when stashing).  While properties can be overloaded via System Properties, this can get cumbersome with a lot of local overrides.  Ideally, I should be able to have a local.properties file (optionally), that overrides config settings, but something I should never have to check into my repo.

johngmyers:
Please file this in the internal ticket system if you still want to move this forward.

PR #669, opened 2018-07-24
#668 Pass requireExplicitBindings flag down to ConfigurationValidator

gdasbach:
#668: Passed the requireExplicitBindings flag down to Configuration Validator to switch between registering unused properties as errors or warnings

PR #671, opened 2018-07-24
#670: Added optional local-config file

gdasbach:
Added additional property for local-config, where developer can set their local configuration property file to override settings in the repo config.properties file

Can be set at launch with -Dlocal-config=etc/local.properties

### DIFF
--- a/configuration/src/main/java/com/proofpoint/configuration/ConfigurationFactoryBuilder.java
+++ b/configuration/src/main/java/com/proofpoint/configuration/ConfigurationFactoryBuilder.java
@@ -20,10 +20,7 @@ import com.google.inject.spi.Message;
 import com.proofpoint.configuration.Problems.Monitor;
 
 import javax.annotation.Nullable;
-import java.io.FileInputStream;
-import java.io.IOException;
-import java.io.InputStreamReader;
-import java.io.Reader;
+import java.io.*;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


gdasbach:
Looks like this is exactly the same CI problem seen on #669.  Rebuild fixed that check, so I'll try the same fix again.

configuration/src/main/java/com/proofpoint/configuration/ConfigurationFactoryBuilder.java
```diff
@@ -20,10 +20,7 @@
 import com.proofpoint.configuration.Problems.Monitor;
 
 import javax.annotation.Nullable;
-import java.io.FileInputStream;
-import java.io.IOException;
-import java.io.InputStreamReader;
-import java.io.Reader;
+import java.io.*;
```
johngmyers: Import the Proofpoint code style into your IDE. Wildcard imports are not permitted.

gdasbach: Fixed.  To answer the other question, in a large team, the project config.properties file is edited quite frequently with new settings for feature implemented by developers.  It's therefore fairly common for a merge with upstream to change values in that file.  If we're using a separate config.properties, we don't see those changes unless we manually port the changes into our local copy.  If we edit the config.properties to contain our local overrides, we frequently have merge problems.  This solution allows us to maintain local overrides in a separate file while still reading configuration from the default config.properties file.

gdasbach:
Left this comment on the previous change request, but haven't heard anything in response, so copying to this discussion.

> To answer the other question, in a large team, the project config.properties file is edited quite frequently with new settings for features implemented by developers. It's therefore fairly common for a merge with upstream to change values in that file. If we're using a separate config.properties, we don't see those changes unless we manually port the changes into our local copy. If we edit the config.properties to contain our local overrides, we frequently have merge problems. This solution allows us to maintain local overrides in a separate file while still reading configuration from the default config.properties file.

johngmyers:
Closing for now per offline conversation.

PR #689, opened 2018-08-27
Merge airlift 2018 part 5

johngmyers:


johngmyers:
Yes, that was an accident. Fixed.

PR #690, opened 2018-08-29
Add Sealights Docker support to library pom

johngmyers:


library/pom.xml
```diff
@@ -1634,5 +1634,55 @@
...
+                                                        <outputDirectory>lib/</outputDirectory>
+                                                    </file>
+                                                    <file>
+                                                        <source>buildSessionId.txt</source>
+                                                        <destName>buildSessionId.txt</destName>
+                                                        <outputDirectory>lib/</outputDirectory>
+                                                    </file>
+                                                    <file>
+                                                        <source>${SL_TOKEN_FILE}</source>
+                                                        <destName>sl-token.txt</destName>
```
rajatjindal: We need fileMode property too

library/pom.xml
```diff
@@ -1634,5 +1634,55 @@
...
+                            <images>
+                                <image>
+                                    <name>${project.docker.project}/${project.docker.name}:${project.version}</name>
+                                    <build>
+                                        <assembly>
+                                            <inline>
+                                                <files>
+                                                    <file>
+                                                        <source>target/sealights/sl-test-listener.jar</source>
+                                                        <destName>sl-test-listener.jar</destName>
```
rajatjindal: Incase of pom.xml inside a sub directory, the location of this file is ../target..... is it possible to figure that out automatically

PR #693, opened 2018-08-30
Start migrating the bindConfig() DSL to ConfigBinder

johngmyers:


johngmyers:
I wonder if we want to change the EDSL to be more like:

```java
configBinder(binder).bind(FooConfig.class)
   .annotatedWith(Red.class)
   .prefixedWith("prefix");
```

gwittel:
Using `bind` as the entry point would be more consistent with the other binders.

Was it intentional to not mark the `ConfigurationModule.bind()` method as deprecated?

johngmyers:
I want to delay deprecating `ConfigurationModule.bindConfig()`  until `ConfigBinder` has existed for a while. It's currently used everywhere.

johngmyers:
Will resubmit with changed EDSL.

PR #697, opened 2018-09-01
Create ConfigBinder

johngmyers:


gwittel:
Looks good.  Should update the sample project as well.

johngmyers:
The sample project is updated in the second commit.

PR #711, opened 2018-10-07
Add support for binding configuration defaults

johngmyers:


johngmyers:
Discovered this is not working when used from a private binder, as is done for the `HttpClientConfig` used under a `BalancingHttpClient`.

PR #712, opened 2018-10-08
Upgrade to Jetty 9.4

johngmyers:


http-client/src/test/java/com/proofpoint/http/client/AbstractHttpClientTest.java
```diff
@@ -1313,6 +1313,11 @@ public void testHandlesUndeclaredThrowable()
     public void testResponseHandlerThrowsBeforeReadingBody()
             throws Exception
     {
+        if (createClientConfig().isHttp2Enabled()) {
+            // To difficult to test with HTTP/2
```
gwittel: Nit - `To -> Too`

I was going to suggest SkipException but it looks like they're broken in TestNG right now.

http-client/src/test/java/com/proofpoint/http/client/AbstractHttpClientTest.java
```diff
@@ -1752,4 +1778,20 @@ private static boolean isConnectTimeout(Throwable t)
     {
         return (E) t;
     }
+
+    // TODO: remove this when fixed in Jetty
+    private static class HackGzipHandler
```
gwittel: Did Jetty have a regression or are we re-introducing this change to keep sync with airlift?

johngmyers: Jetty since fixed that, I believe. Pushed a commit to remove.

PR #719, opened 2018-10-23
HTTP/2 server under TLS in Java 11

johngmyers:


http-server/src/main/java/com/proofpoint/http/server/HttpServer.java
```diff
@@ -391,6 +368,46 @@ protected void runJob(Runnable job)
...
+        sslContextFactory.setKeyStorePassword(config.getKeystorePassword());
+        sslContextFactory.setExcludeProtocols();
+        sslContextFactory.setIncludeProtocols(ENABLED_PROTOCOLS);
+        sslContextFactory.setExcludeCipherSuites();
+        sslContextFactory.setIncludeCipherSuites(ENABLED_CIPHERS);
+        sslContextFactory.setCipherComparator(Ordering.explicit("", ENABLED_CIPHERS));
+
+        List<ConnectionFactory> connectionFactories = new ArrayList<>();
+        connectionFactories.add(new SslConnectionFactory(sslContextFactory, isJava8 ? "http/1.1" : "alpn"));
+        if (!isJava8) {
```
gwittel: This is implicitly enabling HTTP2 for Java 9, 10.  Was the intent 11+ only, or 9+ only? (either code change or readme change)

PR #720, opened 2018-10-24
Adjust JVM flags for launcher and Surefire

johngmyers:


library/pom.xml
```diff
@@ -243,7 +243,7 @@
                             -Dfile.encoding=${project.build.sourceEncoding}
                             -Xmx${platform.test.jvmsize}
                             -Xms${platform.test.jvmsize}
-                            -XX:OnOutOfMemoryError="kill -9 %p"
+                            -XX:+ExitOnOutOfMemoryError
```
gwittel: We should put a note that this release will require 8u101+ (or 8u92+ if on the  PSU series).

PR #722, opened 2018-10-26
Library upgrades

johngmyers:


jmx-http/pom.xml
```diff
@@ -76,8 +76,8 @@
 
         <dependency>
             <groupId>com.google.auto.value</groupId>
-            <artifactId>auto-value</artifactId>
-            <scope>provided</scope>
```
epavlov1: Can provided scope cause an issue ?

johngmyers: This is removing auto-value in provided scope, replacing it with auto-value-annotations as optional (since those are just compile-time annotations). auto-value is now pulled in the annotationProcessorPaths section of the maven-compiler-plugin configuration.

PR #737, opened 2018-11-30
Fix duplicate metric issue

johngmyers:


johngmyers:
Refactored this to use inheritance so that we don't have to make an incompatible API change.

PR #747, opened 2019-02-22
use proofpoint/platform-zulu-8 image.

rajatjindal:


library/pom.xml
```diff
@@ -39,7 +39,7 @@
         <project.rpm.username>${project.artifactId}</project.rpm.username>
         <project.docker.project>dev-docker-local</project.docker.project>
         <project.docker.name>%a</project.docker.name>
-        <project.docker.from>docker.io/azul/zulu-openjdk-debian:8u192</project.docker.from>
+        <project.docker.from>proofpoint/platform-zulu-8:1.0</project.docker.from>
```
johngmyers: Shouldn't we also change the <exec> section in the "docker" profile to invoke through dumb-init?

We'll have to add to CHANGES the requirement that the base image have a dumb-init.

PR #773, opened 2019-10-01
Upgrade Jetty to 9.4.21

johngmyers:


johngmyers:
Tests are failing due to what appears to be a regression in Jetty. Needs more investigation.

PR #778, opened 2019-11-02
Don't override audit log path in launcher

johngmyers:
Some use cases for changing the path within a Docker container.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,10 @@
+Platform 2.17
+
+* Audit Logging
+
+  - The default value for "audit.log.path" has changed to "var/log/audit.log".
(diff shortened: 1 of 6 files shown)


CHANGES
```diff
@@ -1,3 +1,10 @@
+Platform 2.17
+
+* Audit Logging
+
+  - The default value for "audit.log.path" has changed to "var/log/audit.log".
```
gwittel: Probably useful to clarify that audit logging is now enabled by default (vs implying this in the statement about disabling audit logs).

johngmyers: Previously audit logging was enabled by launcher. But I suppose it is changing for in-IDE.

Issue #781, opened 2019-11-25
When using .build-docker the build instuctions for building the docker image gets merged into any other containers you are using


JasonQuinn:
We have additional executions in the docker-maven-plugin which run in the pre-integrarion-test phase to start up db's with the way platform currently works the build configuration gets merged into these containers and then gets pushed to the repo breaking thinks like library/postgres:11 for all other users as well. We also have the problem where because the configuration is set for all executions the pre-integration test step tries to start up the container before it is built and pushed.
If the configuration for the image
<name>${project.docker.project}/${project.docker.name}:${project.version}</name> was moved to the build/push execution then this would not happen.

Survival26:
> We have additional executions in the docker-maven-plugin which run in the pre-integrarion-test phase to start up db's with the way platform currently works the build configuration gets merged into these containers and then gets pushed to the repo breaking thinks like library/postgres:11 for all other users as well. We also have the problem where because the configuration is set for all executions the pre-integration test step tries to start up the container before it is built and pushed.
> If the configuration for the image
> <name>${project.docker.project}/${project.docker.name}:${project.version}</name> was moved to the build/push execution then this would not happen.

d84cc04b5d7f82f10fd50e7c8bfd903b9d33cb4a

PR #782, opened 2019-11-25
#781 - Moving docker image configuration to the build/push execution

JasonQuinn:
Im not sure if changing to use the default-cli id will cause any issues with the non C15 pipelines, but it has to be default-cli for c15 since its uses maven docker:push to push the image so the default id wouldnt work in this case. Should we maybe have the same configurstion for both default and default-cli?


JasonQuinn:
Noticed I never rebased the changes into my fork, going to close this and make a new one with cleaner commits

PR #784, opened 2019-12-12
Remove deprecated APIs

johngmyers:


http-client/src/main/java/com/proofpoint/http/client/HttpClientModule.java
```diff
@@ -104,15 +77,6 @@ public final void configure(Binder binder)
         }
     }
 
-    /**
-     * @deprecated Will no longer be public
-     */
-    @Deprecated
-    public void addAlias(Class<? extends Annotation> alias)
```
gwittel: Looks like this is still referenced in `HttpClientBindingBuilder` causing compilation to fail.

PR #792, opened 2020-02-24
Allow disabling of OverrideMethodFilter

glent:


jaxrs/src/main/java/com/proofpoint/jaxrs/JaxrsConfig.java
```diff
@@ -51,4 +52,18 @@ public boolean isPreload()
     {
         return preload;
     }
+
+    @Deprecated
```
johngmyers: I don't think we should start off deprecating this. We would deprecate after we change the default.

jaxrs/src/main/java/com/proofpoint/jaxrs/JaxrsModule.java
```diff
@@ -196,6 +194,8 @@ public void configure(Binder binder)
             jaxrsBinder(binder).bindAdmin(OverrideMethodFilter.class);
             jaxrsBinder(binder).bindAdmin(ThreadDumpResource.class);
 
+            bindConfig(binder).bind(JaxrsConfig.class);
```
johngmyers: I think I'd prefer unconditionally removing `OverrideMethodFilter` from the admin port. There's unlikely to be a use case for it there. Then we wouldn't have to pull the other options into the admin-only module or split the options into two classes.

jaxrs/src/main/java/com/proofpoint/jaxrs/OverrideMethodFilter.java
```diff
@@ -47,22 +50,33 @@
      */
     public static final String METHOD_PARAM = "_method";
 
+    private final boolean filterEnabled;
+
+    @Inject
+    public OverrideMethodFilter(JaxrsConfig config)
+    {
+        requireNonNull(config, "jaxrsConfig is null");
+        filterEnabled = config.isOverrideMethodFilter();
```
johngmyers: `JaxrsModule` should use `ConditionalModule.installModuleIf()` to decide whether or not to install this module. The pattern of having a filter that perhaps never does anything is brittle.

jaxrs/src/test/java/com/proofpoint/jaxrs/TestOverrideMethodFilter.java
```diff
@@ -122,9 +122,25 @@ public void testHeaderHasPrecedenceOverQueryParam()
...
+        ContainerRequest request = new ContainerRequest(
+                URI.create("http://www.example.com/"),
+                URI.create("http://www.example.com/?_method=PUT"),
+                "POST",
+                new MockSecurityContext(),
+                new MockPropertiesDelegate());
+        request.header(OverrideMethodFilter.HEADER, "DELETE");
+
+        filter.filter(request);
+        assertEqualsIgnoreCase(request.getMethod(), "POST");
```
johngmyers: This is testing the implementation, not the API. A better test would be in `TestJaxrsModule` and would `bootstrapTest()` a server with the config property set and then send a request in through its test client.

jaxrs/src/main/java/com/proofpoint/jaxrs/JaxrsModule.java
```diff
@@ -89,7 +92,12 @@ public void configure(Binder binder)
         jaxrsBinder(binder).bind(SmileMapper.class);
         jaxrsBinder(binder).bind(ParsingExceptionMapper.class);
         jaxrsBinder(binder).bind(QueryParamExceptionMapper.class);
-        jaxrsBinder(binder).bind(OverrideMethodFilter.class);
+        newOptionalBinder(binder, Key.get(Ticker.class, JaxrsTicker.class))
+                .setDefault().toInstance(Ticker.systemTicker());
```
johngmyers: Why this change?

glent: It seems Modules.override doesn't work with AbstractConfigurationAwareModule (got an NPE on buildConfigObject). This change enables the same results (being able to replace the ticker in tests) without requiring a module override. See TestTimingFilter for the related change.

johngmyers:
@glent could you rebase that?

PR #795, opened 2020-03-16
Support config files in JSON format

johngmyers:


configuration/src/test/java/com/proofpoint/configuration/TestPropertiesBuilder.java
```diff
@@ -0,0 +1,133 @@
+/*
+ * Copyright 2010 Proofpoint, Inc.
```
gwittel: 2020?

johngmyers: Ah, had copied `TestConfigurationFactoryBuilder` to start.

PR #802, opened 2020-04-20
Platform 504

Shweta-Deshpande:
Creating this PR to check on the newly added sonar profile.

### DIFF
--- a/library/pom.xml
+++ b/library/pom.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <groupId>com.proofpoint.platform</groupId>
     <artifactId>library</artifactId>
@@ -93,6 +94,8 @@
         <dep.jackson.version>2.10.0</dep.jackson.version>
         <dep.jersey.version>2.27</dep.jersey.version>
         <dep.jmh.version>1.13</dep.jmh.version>
+
+        <platform.jacoco.merged-data-file>${project.build.directory}/merged.exec</platform.jacoco.merged-data-file>
     </properties>
 
     <scm>
@@ -147,7 +150,7 @@
                         <fail>${platform.check.fail-enforcer}</fail>
                         <failFast>false</failFast>
                         <rules>
-                            <requirePluginVersions />
+                            <requirePluginVersions/>
                             <bannedDependencies>
                                 <excludes>
                                     <!-- clashes with commons-logging:commons-logging -->
@@ -180,10 +183,11 @@
                                     <!-- whitelist the well numbered guava releases -->
                                     <include>com.google.guava:guava:[10.0.1,)</include>
                                     <!-- newer versions of junit do not repackage hamcrest -->
-                                    <include>junit:junit:[4.11,)</include>                                </includes>
+                                    <include>junit:junit:[4.11,)</include>
+                                </includes>
                             </bannedDependencies>
-                            <banDuplicatePomDependencyVersions />
-                            <requireUpperBoundDeps />
+                            <banDuplicatePomDependencyVersions/>
+                            <requireUpperBoundDeps/>
                             <requireMavenVersion>
                                 <version>3.2.3</version>
                             </requireMavenVersion>
@@ -467,6 +471,52 @@
                             <goals>
                                 <goal>report</goal>
                             </goals>
+                        </execution>
+                        <execution>
+                            <id>pre-integration-test</id>
+                            <phase>pre-integration-test</phase>
+                            <goals>
+                                <goal>prepare-agent-integration</goal>
+                            </goals>
+                            <configuration>
+                                <propertyName>failsafeArgLine</propertyName>
+                            </configuration>
+                        </execution>
+                        <execution>
+                            <id>post-integration-test</id>
+                            <phase>post-integration-test</phase>
+                            <goals>
+                                <goal>report-integration</goal>
+                            </goals>
+                        </execution>
+                        <execution>
+                            <id>merge-unit-and-integration</id>
+                            <phase>post-integration-test</phase>
+                            <goals>
+                                <goal>merge</goal>
+                            </goals>
+                            <configuration>
+                                <fileSets>
+                                    <fileSet>
+                                        <directory>${project.build.directory}</directory>
+                                        <includes>
+                                            <include>*.exec</include>
+                                        </includes>
+                                    </fileSet>
+                                </fileSets>
+                                <destFile>${platform.jacoco.merged-data-file}</destFile>
+                            </configuration>
+                        </execution>
+                        <execution>
+                            <id>create-merged-report</id>
+                            <phase>post-integration-test</phase>
+                            <goals>
+                                <goal>report</goal>
+                            </goals>
+                            <configuration>
+                                <dataFile>${platform.jacoco.merged-data-file}</dataFile>
+                                <outputDirectory>${project.reporting.outputDirectory}/jacoco-merged</outputDirectory>
+                            </configuration>
                         </execution>
                     </executions>
                     <configuration>
@@ -1438,7 +1488,7 @@
                     <plugin>
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-dependency-plugin</artifactId>
-            			<version>3.1.2</version>
+                        <version>3.1.2</version>
                         <executions>
                             <execution>
                                 <id>unpack launcher</id>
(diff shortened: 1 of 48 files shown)


library/pom.xml
```diff
@@ -94,6 +94,8 @@
         <dep.jackson.version>2.10.0</dep.jackson.version>
         <dep.jersey.version>2.27</dep.jersey.version>
         <dep.jmh.version>1.13</dep.jmh.version>
+
+        <jacoco-merged-report>${project.build.directory}/merged.exec</jacoco-merged-report>
```
johngmyers, line 98: Shouldn't this use ${project.reporting.outputDirectory}?

library/pom.xml
```diff
@@ -244,30 +246,6 @@
                     </configuration>
                 </plugin>
 
-                <plugin>
-                    <groupId>org.apache.maven.plugins</groupId>
-                    <artifactId>maven-surefire-plugin</artifactId>
```
johngmyers, line 249: Why are you moving the maven-surefire-plugin config? This makes things harder to review.

library/pom.xml
```diff
@@ -462,6 +440,9 @@
                             <goals>
                                 <goal>prepare-agent</goal>
                             </goals>
+                            <configuration>
+                                <propertyName>surefireArgLine</propertyName>
```
johngmyers, line 444: Why do you need to change this from the default?

library/pom.xml
```diff
@@ -470,9 +451,82 @@
...
+                        </execution>
+                        <execution>
+                            <id>post-integration-test</id>
+                            <phase>post-integration-test</phase>
+                            <goals>
+                                <goal>report-integration</goal>
+                            </goals>
+                        </execution>
+                        <execution>
+                            <id>merge-unit-and-integration</id>
```
johngmyers, line 476: This will cause everyone to get a merged report. Is that really what everyone will want?

library/pom.xml
```diff
@@ -470,9 +451,82 @@
...
                     </executions>
+                </plugin>
+
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-surefire-plugin</artifactId>
+                    <version>2.22.2</version>
                     <configuration>
-                        <skip>${platform.check.skip-jacoco}</skip>
+                        <forkCount>1</forkCount>
```
johngmyers, line 512: Why are you explicitly setting the default value?

library/pom.xml
```diff
@@ -470,9 +451,82 @@
...
+                        <reuseForks>true</reuseForks>
```
johngmyers, line 513: Why explicitly setting the default value?

library/pom.xml
```diff
@@ -1714,5 +1768,41 @@
...
+            </activation>
+            <properties>
+                <sonar.branch.name>${env.BRANCH_NAME}</sonar.branch.name>
+            </properties>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.jacoco</groupId>
+                        <artifactId>jacoco-maven-plugin</artifactId>
+                        <version>0.8.5</version>
```
johngmyers, line 1787: Why the duplicate version specification here?

library/pom.xml
```diff
@@ -1714,5 +1768,41 @@
...
+                    </plugin>
+                    <plugin>
+                        <groupId>org.sonarsource.scanner.maven</groupId>
+                        <artifactId>sonar-maven-plugin</artifactId>
+                        <version>3.7.0.1746</version>
```
johngmyers, line 1792: Why the duplicate version specification here?

library/pom.xml
johngmyers: Also this doesn't match the naming conventions. It should probably be under "project.". But why would one want the jacoco merged report to go somewhere other than project.reporting.outputDirectory?

Shweta-Deshpande: This is not the actual report's path. It is the path to the output file for execution data. (.exec). I can change the name to the property to jacoco-merged-execution-data.

library/pom.xml
Shweta-Deshpande: I didn't have any specific reason to move it. Just happened as part of my implementation flow.

library/pom.xml
johngmyers: But it really shouldn't go to the top-level build directory, where "mvn clean" won't remove it. Shouldn't it go into into the project.reporting.outputDirectory?

And "jacoco-merged-execution-data" doesn't match the naming conventions either. You'll notice the other properties here don't have top-level names.

library/pom.xml
johngmyers: Could you move it back and squash?

library/pom.xml
Shweta-Deshpande: Need to change this because I created 2 properties now. One for unit tests using maven surefire called "surefireArgLine" and the other for integration tests using maven failsafe called "failsafeArgLine". Projects using failsafe plugins can simply add "failsafeArgLine" to their configuration.

library/pom.xml
Shweta-Deshpande: Yes this will create a merged report for everyone. But using this report for SQ analysis is up to the project requirements. If a project requires to analysis for UT and IT combined they can use this report by adding to their maven build command: 
-Dsonar.coverage.jacoco.xmlReportPaths=target/site/jacoco-merged/jacoco.xml

library/pom.xml
Shweta-Deshpande: As per this article:
http://www.ffbit.com/blog/2014/05/21/skipping-jacoco-execution-due-to-missing-execution-data-file/

library/pom.xml
Shweta-Deshpande: As per this article:
http://www.ffbit.com/blog/2014/05/21/skipping-jacoco-execution-due-to-missing-execution-data-file/

library/pom.xml
Shweta-Deshpande: I thought we decided to go the default way as much as possible. And as per the jacoco documentation:
Path to the output file for execution data.
Default value is: ${project.build.directory}/jacoco.exec.
User property is: jacoco.destFile.

I'm sorry I don't know what the naming convention is. Should it be <platform.jacoco.merged.execution.data> instead?

library/pom.xml
Shweta-Deshpande: Ok.

library/pom.xml
Shweta-Deshpande: I didn't notice the version specified elsewhere.

library/pom.xml
Shweta-Deshpande: I didn't notice the version specified elsewhere.

library/pom.xml
johngmyers: A lower impact change would be to leave surefire with `argLine` and introduce `failsafeArgLine` with the addition of failsafe

library/pom.xml
johngmyers: But `1` is the default value of `forkCount`. Have you seen any problems? The article is a bit fuzzy, but looks like it is saying "if you unintentionally change these values you could cause a problem".

library/pom.xml
johngmyers: Line 482

library/pom.xml
Shweta-Deshpande: Oh I see it now. Will remove the version from here.

library/pom.xml
Shweta-Deshpande: Oh I see it now. Will remove the version from here.

library/pom.xml
Shweta-Deshpande: Yes I saw an error during my development. "Skipping JaCoCo Execution Due to Missing Execution Data File". Thats when I googled and found this article.

I will test this again without the explicit setting of forkCount to 1 and remove it if I don't see any errors.

library/pom.xml
Shweta-Deshpande: Done.

library/pom.xml
Shweta-Deshpande: Renamed  jacoco-merged-execution-data to platform.jacoco.merged.execution.data.

library/pom.xml
```diff
@@ -470,10 +477,53 @@
...
+                            </goals>
+                            <configuration>
+                                <dataFile>${platform.jacoco.merged.execution.data}</dataFile>
+                                <outputDirectory>${project.reporting.outputDirectory}/jacoco-merged</outputDirectory>
+                            </configuration>
+                        </execution>
                     </executions>
-                    <configuration>
-                        <skip>${platform.check.skip-jacoco}</skip>
-                    </configuration>
```
johngmyers, line 476: We should leave this in.

library/pom.xml
johngmyers: I had thought we were not going to enable the integration tests in Platform as there were so few projects using them. The cost is the time to do the extra merging and report for the vast majority of projects that don't have maven integration tests.

library/pom.xml
```diff
@@ -94,6 +94,8 @@
         <dep.jackson.version>2.10.0</dep.jackson.version>
         <dep.jersey.version>2.27</dep.jersey.version>
         <dep.jmh.version>1.13</dep.jmh.version>
+
+        <platform.jacoco.merged.execution.data>${project.build.directory}/merged.exec</platform.jacoco.merged.execution.data>
```
johngmyers, line 98: I see why we want to change the default used by jacoco:merge's destFile. I'm not quite clear why we're exposing this as a property to change.

If we do want a property, then I think `platform.jacoco.merged-data-file` would be a reasonable name.

library/pom.xml
Shweta-Deshpande: If a project does not have integration tests the merged report will essentially be similar to the UT report. There will not be any extra merging. I think this is a good to have since I have seen few projects which prefer having more ITs than UTs and look at coverage from a combined perspective.

library/pom.xml
Shweta-Deshpande: I created a property since this path is used in 2 places. I treated it as a variable declaration.

library/pom.xml
Shweta-Deshpande: By having the merged report we are not enabling the integration tests. To enable the ITs developers will have to include the failsafe plugin in their poms and then use the failsafeArgLine in it.

library/pom.xml
Shweta-Deshpande: Ok.

library/pom.xml
Shweta-Deshpande: Done

library/pom.xml
```diff
@@ -456,6 +458,9 @@
                     <groupId>org.jacoco</groupId>
                     <artifactId>jacoco-maven-plugin</artifactId>
                     <version>0.8.5</version>
+                    <configuration>
+                        <skip>${platform.check.skip-jacoco}</skip>
+                    </configuration>
```
johngmyers, line 463: Please put this back where it used to be.

Shweta-Deshpande: Done

PR #803, opened 2020-04-22
Update platform-zulu from 1.5 to 1.6

johnmccabe:
Base image includes the following changes:
- update zulu to 8u252-8.46.0.19 and 11.0.7-11.39.15
- zulu image is now based on bionic
- also pins to specific release for reproducability

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,11 @@
+Platform 2.28
+
+* Library Upgrades
+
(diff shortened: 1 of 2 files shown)


johnmccabe:
I'd pulled from my fork rather than upstream, rebasing and will push updated changes

johnmccabe:
Updated description to point out that base image is Debian buster, and not bionic as originally reported.

PR #804, opened 2020-05-02
Move StartLimitIntervalSecs setting back to [Unit] and rename

johngmyers:


CHANGES
```diff
@@ -1,3 +1,10 @@
+Platform 2.29
+
+* Launcher
+
+  We moved the StartLimitIntervalSecs setting back from [Service] to [Unit] and
```
gwittel, line 5: from unit to service?

johngmyers: Huh, looks like I got the previous release note backwards.

johngmyers: Fixed

PR #809, opened 2020-06-18
SonarQube PR analysis integration

HONTAKYU:
LEGO-664, LEGO-624
Adding a new (similar) maven profile to enable SonarQube PR analysis.

CHANGE_ID will be available only on PR, therefore a different set of property will be executed. This will give developers more insights of the quality of PR, and even make them aware of this on the Bitbucket side.

### DIFF
--- a/library/pom.xml
+++ b/library/pom.xml
@@ -1774,7 +1774,7 @@
             </activation>
             <properties>
                 <sonar.branch.name>${env.BRANCH_NAME}</sonar.branch.name>
-                <sonar.projectKey>${env.SONAR_PROJECT_KET}</sonar.projectKey>
+                <sonar.projectKey>${env.SONAR_PROJECT_KEY}</sonar.projectKey>
             </properties>
             <build>
                 <plugins>
@@ -1813,7 +1813,7 @@
                 <sonar.pullrequest.key>${env.CHANGE_ID}</sonar.pullrequest.key>
                 <sonar.pullrequest.branch>${env.CHANGE_BRANCH}</sonar.pullrequest.branch>
                 <sonar.pullrequest.base>${env.CHANGE_TARGET}</sonar.pullrequest.base>
-                <sonar.projectKey>${env.SONAR_PROJECT_KET}</sonar.projectKey>
+                <sonar.projectKey>${env.SONAR_PROJECT_KEY}</sonar.projectKey>
             </properties>
             <build>
                 <plugins>
--- a/library/pom.xml
+++ b/library/pom.xml
@@ -1763,14 +1763,56 @@
         </profile>
 
         <profile>
-            <id>build-sonar</id>
+            <id>sonar-branch</id>
             <activation>
                 <file>
                     <exists>.build-sonar</exists>
                 </file>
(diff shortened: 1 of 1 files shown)


library/pom.xml
```diff
@@ -1763,11 +1763,14 @@
...
             <activation>
                 <file>
                     <exists>.build-sonar</exists>
                 </file>
+                <property>
+                    <name>!env.CHANGE_ID</name>
+                </property>
             </activation>
             <properties>
                 <sonar.branch.name>${env.BRANCH_NAME}</sonar.branch.name>
```
gwittel, line 1776: We should probably change both cases of `sonar.projectKey` to use a neutral environment variable like `SONAR_PROJECT_KEY`.  This lets users who don't use bitbucket run it.

The build system can set it to the bitbucket specific value you have in this revision.  It also means if the format changes, people don't have to upgrade platform.

HONTAKYU: env.BITBUCKET_PROJECT_KEY and env.BITBUCKET_REPO_NAME are actually Bitbucket-independent environment variables. The name is misleading, they are actually coming from git config. Details can be found: https://hq-stash.corp.proofpoint.com/projects/DEL/repos/k8s-pipeline/commits/087d4d0798fc59a4bc1157dfd29bb928cd0cb065#vars/initPipeline.groovy

gwittel: Got it.  Either way, I still think it should be a single parameter value so we're decoupling build from the format here.

HONTAKYU: Sure, I can update other related repos to reflect this change.

Shweta-Deshpande: I am not sure if we can decouple the scm (source code management systems) type from the build. A project key and repo name distinction is not used by all scm types. This usage if env. BITBUCKET_* env variables is relevant to our use case since we expect the git config to return a clone url for bitbucket. 
Here is the backend code for setting these vars:

    ORIGIN = sh (
            script: "git config --get remote.origin.url",
            returnStdout: true
    ).trim()

    items = ORIGIN.split("/")
    reponame = items[-1].replaceAll(/\.git$/, "")
    projectkey = items[-2].toLowerCase()

This assumes urls of the form:
ssh://git@hq-stash.corp.proofpoint.com:7999/tqa/cd-poc.git
and extracts projectkey = tqa
reponame = cd-poc

This one of the reasons I wished we could go with the sonar default for setting the project key which is <groupId>:<artifactId> from the pom.xml. 
But we have a hard set requirement to maintain it as porject_key:repo_name.

gwittel: The point is to have:

```
<project.key>${env.SONAR_PROJECT_KEY}</project.key>
```

So our build system can set:

```
SONAR_PROJECT_KEY="${BITBUCKET_PROJECT_KEY}:${BITBUCKET_REPO_NAME}"
```

This makes the platform release independent of underlying build/git internal organization.

Shweta-Deshpande: Ah got it!

HONTAKYU: Done, please review

PR #814, opened 2020-07-02
Lego 802

Shweta-Deshpande:


library/pom.xml
```diff
@@ -1803,12 +1812,12 @@
                 </property>
             </activation>
             <properties>
-                <sonar.branch.name>${env.BRANCH_NAME}</sonar.branch.name>
+                <sonar.branch.name>${GIT_BRANCH}</sonar.branch.name>
```
johngmyers, line 1815: Shouldn't this be `env.GIT_BRANCH`?

johngmyers:
Could you rebase and squash?

library/pom.xml
Shweta-Deshpande: It does not work with the "env." prefix.

Shweta-Deshpande:
Creating a new PR.

PR #822, opened 2020-10-05
Defer creation of HTTP listener sockets to @AcceptRequests

johngmyers:


gwittel (approved):
LGTM

http-server/src/main/java/com/proofpoint/http/server/HttpServerInfo.java
```diff
@@ -165,4 +178,116 @@ private static ServerSocketChannel createChannel(InetAddress address, int port,
...
+
+    static class LazyChannel implements ChannelHolder
+    {
+        private final String scheme;
+        private final String hostname;
+        private final String externalHostname;
+        private final InetAddress address;
+        private final int port;
+        private final int acceptQueueSize;
+        private boolean isAdmin;
```
therontock, line 220: shouldn't admin be final as well?

http-server/src/main/java/com/proofpoint/http/server/HttpServerInfo.java
```diff
@@ -165,4 +178,116 @@ private static ServerSocketChannel createChannel(InetAddress address, int port,
...
+            synchronized (this) {
+                if (channel == null) {
+                    try {
+                        ServerSocketChannel newChannel = ServerSocketChannel.open();
+                        newChannel.socket().setReuseAddress(true);
+                        newChannel.socket().bind(new InetSocketAddress(address, port), acceptQueueSize);
+                        channel = newChannel;
+                    }
+                    catch (IOException e) {
+                        throw new UncheckedIOException(e);
```
therontock, line 252: Should call newChannel.close() if newChannel != null.

johngmyers: `newChannel` is out of scope here. While we could do a nested try block, that's getting overly complicated for what is in effect going to shut down the process.

PR #823, opened 2020-10-06
Reduce default accept queue size to 10

johngmyers:


gwittel (approved):
LGTM

nr17 (approved):
What's the default number of worker threads handling the connections? If we are reducing it to 10, hopefully the workers aren't going to be starved of work and thereby reducing the effective throughput.

johngmyers:
This limits the number of new connections that can queue before having a worker thread call accept().

http-server/src/main/java/com/proofpoint/http/server/HttpServerConfig.java
```diff
@@ -61,9 +61,9 @@
       The OS can truncate the accept queue size to the platform defaults.
       E.g. for linux, the effective server listen backlog queue size cannot be set greater
       than /proc/sys/net/core/somaxconn on the system.
-      If you feel the need to increase httpAcceptQueueSize, increase the OS limit first.
+      If you feel the need to increase httpAcceptQueueSize, you might need to increase the OS limit first.
     */
-    private int httpAcceptQueueSize = 128;
+    private int httpAcceptQueueSize = 10;
```
therontock, line 66: What about warning the user if this value is > 128?  It's anyway going to be ignored, and setting it to a large value is inviting the kind of fail-to-make-progress problem that we saw with OMS-4005.

johngmyers: The comment suggests that with an OS configuration change the value could be >128.

therontock: Very few people will see the comment in the code.  But a warning might get someone to notice.

PR #824, opened 2020-10-06
Cleanup and review comment

johngmyers:


gwittel (approved):
LGTM

therontock:
LGTM to me too.  Personally I prefer the approach of declaring the resource to be cleaned up outside the try block and initialized with null.  Then in the catch block you can check if the value is non-null and close it.  I think that's a bit easier to read than lots of nested try blocks.

therontock (approved):

PR #826, opened 2020-11-07
remove creation of service user to remove sh dependency

nr17:
service user and /service/var are assumed to be created in the base image

### DIFF
--- a/library/pom.xml
+++ b/library/pom.xml
@@ -39,7 +39,7 @@
         <project.rpm.username>${project.artifactId}</project.rpm.username>
         <project.docker.project>dev-docker-local</project.docker.project>
         <project.docker.name>%a</project.docker.name>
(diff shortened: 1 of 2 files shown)


library/pom.xml
```diff
@@ -1696,13 +1695,6 @@
                                     <build>
                                         <from>${project.docker.from}</from>
                                         <optimise>true</optimise>
-                                        <runCmds combine.children="append">
-                                            <run>useradd -M -d /service -u ${project.docker.uid} service</run>
-                                            <run>mkdir /service/var</run>
-                                            <run>chown service /service/var</run>
-                                        </runCmds>
-                                        <user>${project.docker.uid}</user>
-                                        <workdir>/service</workdir>
```
johngmyers, line 1705: I think we need to keep the user and workdir configuration.

nr17: restored them.

johngmyers (approved):

PR #833, opened 2020-12-15
Make the JSON request log less verbose

johngmyers:


http-server/src/main/java/com/proofpoint/http/server/HttpRequestEvent.java
```diff
@@ -149,62 +149,61 @@ private HttpRequestEvent(
...
 
-    @JsonProperty
+    @JsonProperty("u")
     public String getRequestUri()
     {
         return requestUri;
     }
 
-    @JsonProperty("username")
+    @JsonProperty
```
nr17, line 183: missing "user"?

gwittel (approved):
LGTM.  Given nature of change it would be good to get at least one other person to take a look.

http-server/src/main/java/com/proofpoint/http/server/HttpRequestEvent.java
```diff
@@ -215,25 +222,25 @@ public long getTimeToLastByte()
         return timeToLastByte;
     }
 
-    @JsonProperty("timeToLastByte")
+    @JsonProperty("tl")
     public Duration getTimeToLastByteDuration() {
         return new Duration(timeToLastByte, MILLISECONDS);
```
therontock, line 227: I've wondered about this format for reporting a time value.  Wouldn't it be easier to process this if it were reported as an int, instead of a string value like "130ms"?  It would also be a few bytes shorter.  :-)

therontock (approved):
I assume this is to reduce usage for splunk?

gwittel:
> I assume this is to reduce usage for splunk?

Yes

glent (approved):

PR #843, opened 2021-03-22
Add message to the json codec exception

nr17:


gwittel:
Was there a particular request/problem driving this change?  Strictly valid data is good, but given the lack of controls over the default codec behavior I'm a little concerned it will have surprise side effects.  In particular:

1. Services not having a way to disable this end to end.  Its a general problem since much of platform relies on the default, inaccessible/nonconfigurable ObjectMapper.
2. The [performance warning](https://fasterxml.github.io/jackson-core/javadoc/2.12/com/fasterxml/jackson/core/JsonParser.Feature.html#STRICT_DUPLICATE_DETECTION) in Jackson docs about this feature:
    >  Note that enabling this feature will incur performance overhead due to having to store and check additional information: this typically adds 20-30% to execution time for basic parsing.
3. Services don't necessarily have control over their data and must sometimes be tolerant of bad data rather than blowing up.

json/src/main/java/com/proofpoint/json/JsonCodec.java
```diff
@@ -30,14 +23,23 @@
...
 import java.util.function.Supplier;
 
+import com.fasterxml.jackson.core.JsonParseException;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.JavaType;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.base.Suppliers;
+import com.google.common.reflect.TypeParameter;
+import com.google.common.reflect.TypeToken;
+import com.proofpoint.json.LengthLimitedWriter.LengthLimitExceededException;
```
johngmyers, line 33: The reordering of imports suggests that the Proofpoint code style hasn't been loaded.

json/src/test/java/com/proofpoint/json/TestJsonCodec.java
```diff
@@ -240,4 +241,16 @@ public void testToJsonWithLengthLimitComplex()
...
+    @Test
+    public void testDetectDuplicateKeys() {
+        JsonCodec<PersonMap> jsonCodec = jsonCodec(PersonMap.class);
+        String json = "{\"id\": 123, \"persons\": {\"A\": {\"name\": \"John\", \"rocks\": \"true\"}, \"A\": {\"name\": \"John\", \"rocks\": \"true\"}}}";
+        Throwable throwable = expectThrows(IllegalArgumentException.class, () -> {
+            PersonMap m = jsonCodec.fromJson(json);
+            System.out.println(m);
+        });
+        System.out.println(throwable);
+        assertTrue(throwable.getMessage().contains("Duplicate field \'A\'"));
```
johngmyers, line 254: Why not use the `expectedExceptions` and `expectedExceptionsMessageRegExp` parameters of `@Test`?

json/src/main/java/com/proofpoint/json/JsonCodec.java
```diff
@@ -30,14 +23,23 @@
...
+
 import static com.fasterxml.jackson.databind.SerializationFeature.INDENT_OUTPUT;
 import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
 
 public class JsonCodec<T>
 {
     private static final Supplier<ObjectMapper> OBJECT_MAPPER_SUPPLIER = Suppliers.memoize(
-            () -> new ObjectMapperProvider().get().enable(INDENT_OUTPUT))::get;
+            () -> new ObjectMapperProvider().get().enable(INDENT_OUTPUT).enable(JsonParser.Feature.STRICT_DUPLICATE_DETECTION))::get;
```
johngmyers, line 42: I don't see why this should be a quirk of JsonCodec. Why isn't it set in ObjectMapperProvider like all the other Features.

johngmyers (changes requested):

nr17:
> Was there a particular request/problem driving this change? Strictly valid data is good, but given the lack of controls over the default codec behavior I'm a little concerned it will have surprise side effects. In particular:
> 
> 1. Services not having a way to disable this end to end.  Its a general problem since much of platform relies on the default, inaccessible/nonconfigurable ObjectMapper.
> 2. The [performance warning](https://fasterxml.github.io/jackson-core/javadoc/2.12/com/fasterxml/jackson/core/JsonParser.Feature.html#STRICT_DUPLICATE_DETECTION) in Jackson docs about this feature:
>    > Note that enabling this feature will incur performance overhead due to having to store and check additional information: this typically adds 20-30% to execution time for basic parsing.
> 3. Services don't necessarily have control over their data and must sometimes be tolerant of bad data rather than blowing up.

The original issue is that if there is a user-error in json, service will not know which config is effective and what other config gets discarded. This can happen with configuration files/input where there are duplicate keys in a potentially large json file. The current json codec behavior is to deduplicate it and present just one out of the duplicates.

There are few options

1. To keep the existing behavior, we could potentially add another way of achieving this where we have a new method : 

```
public JsonCodec<T> withoutDuplicate()
{
    return new JsonCodec<>(mapper.copy().enable(JsonParser.Feature.STRICT_DUPLICATE_DETECTION), type);
}
```
2. Change the default behavior, and provide a method to discard duplicate keys. 

I would assume that people would not unknowingly want to discard duplicate keys. But I can understand the performance issue and the requirement for tolerating bad data.

johngmyers:
I believe the main driver of this is config files, but the issue exists for APIs as well.

RFC 8259 states "When the names within an object are not unique, the behavior of software that receives such an object is unpredictable."

We already disable `DeserializationFeature.ACCEPT_FLOAT_AS_INT` in `ObjectMapperProvider.get()`, so we already have a stance of strictness in parsing.

json/src/main/java/com/proofpoint/json/JsonCodec.java
johngmyers: I think we should be setting `DeserializationFeature.FAIL_ON_READING_DUP_TREE_KEY` instead.

nr17: This doesn't work as it requires us to deserialize the json string/input to jackson json tree first and then to the POJO. Currently we are calling ObjectMapper.readValue(json, javaType) instead of JsonParser.readValueAsTree(). 
If we change, we'll have to do something like this:
```
        JsonParser jsonParser = < ... > // create a new instance with json bytes ? have to look into it. mapper.getJsonFactory().createJsonParser(writeValueAsBytes(pojo));

        JsonNode tree = jsonParser.readValueAsTree();  // --> catch the exception here
        return  jsonParser.getCodec().treeToValue(tree, MyPOJO.class);
```

This might have performance impact.

json/src/test/java/com/proofpoint/json/TestJsonCodec.java
johngmyers: This should use `JsonTester.decodeJson()` instead of using a string constant.

json/src/test/java/com/proofpoint/json/PersonMap.java
```diff
@@ -0,0 +1,37 @@
...
+package com.proofpoint.json;
+
+import java.util.Map;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class PersonMap
+{
+    @JsonProperty
+    private int id;
```
johngmyers, line 25: This is extraneous to the test.

json/src/test/java/com/proofpoint/json/PersonMap.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @JsonProperty
+    private Map<String, Person> persons;
```
johngmyers, line 28: This shows poor practice in deserialization codecs. The object should be immutable. Use a `@JsonCreator` instead.

json/src/main/java/com/proofpoint/json/JsonCodec.java
johngmyers: I've been delving into the databind code and it doesn't look very consistent about when it tests the `FAIL_ON_READING_DUP_TREE_KEY` feature. In particular, `BeanDeserializer` doesn't test it, so it won't be effective for `Person`.

I did find that if the codec is defined as taking an `ImmutableMap` instead of a `Map` then one gets the desired exception. This is because `ImmutableMapBuilder.put()` disallows duplicate keys. Perhaps that would be a better way to approach the problem.

json/src/main/java/com/proofpoint/json/JsonCodec.java
nr17: Can you point me to the code style file. I have one file already loaded and it is called PFPT. I obtained it from scancloud project. The sort order of imports is slightly different in that. https://wiki.proofpoint.com/wiki/display/ENG/Java+Code+Style#JavaCodeStyle-Imports suggests existence of the file: "Code style for IntelliJ is tracked in Git".

json/src/test/java/com/proofpoint/json/TestJsonCodec.java
nr17: decodeJson itself calls codec.fromJson which we want to test here.

johngmyers: The test here uses JSON encoded as a string literal, which is hard to read and maintain. `JsonTester.decodeJson()` has tests represent the JSON they want to test as much more readable POJO. See https://wiki.proofpoint.com/wiki/display/iEng/JSON+serialization+with+Jackson

nr17: I had considered the TestPersonRepresentation example after receiving your comments. Since the json itself is invalid due to duplicate keys, I don't know how to represent that in a readable POJO (hashMap). Do you have any suggestion?

johngmyers: Ah, good point. I wonder how multimaps serialize.

json/src/main/java/com/proofpoint/json/ObjectMapperProvider.java
```diff
@@ -113,6 +113,9 @@ public ObjectMapper get()
         // do not allow converting a float to an integer
         objectMapper.disable(DeserializationFeature.ACCEPT_FLOAT_AS_INT);
 
+        // do not allow duplicate keys
+        objectMapper.enable(DeserializationFeature.FAIL_ON_READING_DUP_TREE_KEY);
```
johngmyers, line 117: Why is this still needed?

nr17: didn't realize that it was working just because of changing from Map to ImmutableMap. Removed it now.

PR #852, opened 2021-04-16
Upgrade Jetty to 9.4.40

johngmyers:
Also, GitHub Actions appears to have broken making connections to the test runner's own public IP address, so adjust the test suite appropriately.


nr17 (approved):
lgtm

PR #861, opened 2021-08-07
Fix or suppress modernizer errors

johngmyers:


pom.xml
```diff
@@ -15,7 +15,7 @@
     </parent>
 
     <properties>
-        <platform.check.fail-modernizer>false</platform.check.fail-modernizer>
```
gwittel, line 18: Minor: We should document this behavior change in the changelog.

johngmyers: That is in the platform pom, not the library pom, so only affects Platform proper.

gwittel: Ah oops, missed that.

gwittel (approved):
LGTM

PR #867, opened 2021-09-20
Upgrade jakarta.el to 4.0.2 for configuration module

jmdacruz:


jmdacruz:
not needed, since version is inherited from dependencyManagement section in the library pom

PR #876, opened 2022-04-01
log a warning when service descriptors are missing

tbye-pfp:
It can be unexpected that a discovered client keeps working when there are no service descriptions in discovery. Log a warning to indicate what is happening.

### DIFF
--- a/discovery/src/main/java/com/proofpoint/discovery/client/ServiceDescriptorsUpdater.java
+++ b/discovery/src/main/java/com/proofpoint/discovery/client/ServiceDescriptorsUpdater.java
@@ -109,14 +109,24 @@ public final class ServiceDescriptorsUpdater
             {
                 Duration delay = null;
 
-                // If we have no current descriptors, or the new descriptors have contents, perform the set.
-                if (serviceDescriptors.get() == null || !newDescriptors.getServiceDescriptors().isEmpty()) {
+                if (newDescriptors.getServiceDescriptors().isEmpty()) {
+                    // If no new service descriptors are available, log a warning.
+                    // Keep any previous service descriptors to provide some robustness for degraded operation.
+                    if (serviceDescriptors.get() == null) {
+                        log.warn("Discovery returned zero available service instances. No previous service " +
+                                "descriptors are available.");
+                    }
+                    else {
+                        log.warn("Discovery returned zero available service instances. Keeping previous set of instances.");
+                    }
+                }
+                else {
+                    // Update with the new descriptors.
                     serviceDescriptors.set(newDescriptors);
                     target.updateServiceDescriptors(newDescriptors.getServiceDescriptors());
                     delay = newDescriptors.getMaxAge();
                 }
-                // else current descriptors are not null (i.e. have previously been set), *and* new descriptors are empty;
-                // therefore do not replace.  a little bit of robustness for degraded operation.
+
                 errorBackOff.success();
 
                 if (delay == null) {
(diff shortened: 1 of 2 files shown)


johngmyers:
Perhaps we should also log when there are no service descriptors and there is no non-empty previous set?

tbye-pfp:
@johngmyers  Good idea. I created two different warning messages for more context on what is happening.

discovery/src/main/java/com/proofpoint/discovery/client/ServiceDescriptorsUpdater.java
```diff
@@ -109,14 +109,21 @@ public void onSuccess(ServiceDescriptors newDescriptors)
             {
                 Duration delay = null;
 
-                // If we have no current descriptors, or the new descriptors have contents, perform the set.
-                if (serviceDescriptors.get() == null || !newDescriptors.getServiceDescriptors().isEmpty()) {
+                if(newDescriptors.getServiceDescriptors().isEmpty()) {
```
johngmyers, line 112: space after `if`

discovery/src/main/java/com/proofpoint/discovery/client/ServiceDescriptorsUpdater.java
```diff
@@ -109,14 +109,22 @@ public void onSuccess(ServiceDescriptors newDescriptors)
...
 
-                // If we have no current descriptors, or the new descriptors have contents, perform the set.
-                if (serviceDescriptors.get() == null || !newDescriptors.getServiceDescriptors().isEmpty()) {
+                if (newDescriptors.getServiceDescriptors().isEmpty()) {
+                    // If no new service descriptors are available, log a warning.
+                    // Keep any previous service descriptors to provide some robustness for degraded operation.
+                    if (serviceDescriptors.get() == null) {
+                        log.warn("Discovery returned zero available service instances. No previous service " +
+                                     "descriptors are available.");
+                    } else {
```
johngmyers, line 118: Proofpoint style puts else on the next line

discovery/src/main/java/com/proofpoint/discovery/client/ServiceDescriptorsUpdater.java
```diff
@@ -109,14 +109,22 @@ public void onSuccess(ServiceDescriptors newDescriptors)
...
+                        log.warn("Discovery returned zero available service instances. Keeping previous set of instances.");
+                    }
+                } else {
```
johngmyers, line 121: style

johngmyers:
Looks good. Please squash.

tbye-pfp:
@johngmyers it's squashed

johngmyers:
@tbye-pfp The unit test needs to be updated

```
[ERROR] Failures: 
2022-04-05T17:43:58.6506799Z [ERROR]   TestHttpServiceBalancerListenerAdapter.testStartedEmpty:99 
2022-04-05T17:43:58.6507280Z Wanted but not invoked:
2022-04-05T17:43:58.6507661Z httpServiceBalancerImpl.updateHttpUris(
2022-04-05T17:43:58.6508043Z     <Capturing argument>
2022-04-05T17:43:58.6508342Z );
2022-04-05T17:43:58.6509476Z -> at com.proofpoint.discovery.client.balancing.TestHttpServiceBalancerListenerAdapter.testStartedEmpty(TestHttpServiceBalancerListenerAdapter.java:99)
2022-04-05T17:43:58.6510265Z Actually, there were zero interactions with this mock.
2022-04-05T17:43:58.6510518Z 
```

johngmyers:
I'm looking into the semantic that unit test is enforcing.

johngmyers:
Both implementations of `ServiceDescriptorsListener` that code currently feeds into initialize themselves to contain an empty `Iterable`, so don't rely on an initial call. A new implementation could, in theory, use such an initial call to know when a Discovery call completed, but it wouldn't get notified if the calls failed instead of returning zero results. So I don't think we have to preserve the semantic the test is verifying.

PR #907, opened 2023-03-08
Upgrade modernizer and jacoco to support Java 17

tomjshore:
When we have a project on Java 17 and the parent pom is platform, we find that modernizer and jacoco do not run correctly. 

For example
```Execution modernizer of goal org.gaul:modernizer-maven-plugin:2.2.0:modernizer failed: Unsupported class file major version 61```

To fix this I have upgraded jacoco and modernizer.

Because I have updated modernizer I had to fix some bits of the code to compile with new modernizer rules.

### DIFF
--- a/discovery/src/main/java/com/proofpoint/discovery/client/testing/StaticServiceSelector.java
+++ b/discovery/src/main/java/com/proofpoint/discovery/client/testing/StaticServiceSelector.java
@@ -17,10 +17,10 @@ package com.proofpoint.discovery.client.testing;
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Iterables;
 import com.proofpoint.discovery.client.ServiceDescriptor;
 import com.proofpoint.discovery.client.ServiceSelector;
 
+import java.util.Collection;
 import java.util.List;
 
 import static com.proofpoint.discovery.client.ServiceSelectorConfig.DEFAULT_POOL;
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: discovery/src/main/java/com/proofpoint/discovery/client/testing/StaticServiceSelector.java [hunk 1]
```diff
@@ -37,11 +37,11 @@ public StaticServiceSelector(ServiceDescriptor... serviceDescriptors)
         this(ImmutableList.copyOf(serviceDescriptors));
     }
 
-    public StaticServiceSelector(Iterable<ServiceDescriptor> serviceDescriptors)
+    public StaticServiceSelector(Collection<ServiceDescriptor> serviceDescriptors)
```
johngmyers: This appears to be an incompatible change.

reporting-prometheus/src/main/java/com/proofpoint/reporting/MetricsResource.java
```diff
@@ -77,7 +78,7 @@ public StreamingOutput getMetrics() {
                         taggedValue.getValueAndTimestamp().getValue().writeMetric(
                                 writer,
                                 entry.getKey(),
-                                Iterables.concat(taggedValue.getTags().entrySet(), instanceTags.entrySet()),
+                                Stream.concat(taggedValue.getTags().entrySet().stream(), instanceTags.entrySet().stream()).collect(Collectors.toList()),
```
johngmyers, line 81: This will cause the tags to be copied to a newly allocated `List` instead of allowing the writer to iterate over the concatination.

johngmyers (changes requested):

reporting-prometheus/src/main/java/com/proofpoint/reporting/MetricsResource.java
tomjshore: Yep ok I have reverted this back and added an exception to this rule in the pom modernizer section.

discovery/src/main/java/com/proofpoint/discovery/client/testing/StaticServiceSelector.java
tomjshore: I was playing around with this in an early commit.  You should see now that the signature is back to Iterable.

tomjshore:
This ticket is ready to be reviewed again.

johngmyers (approved):

PR #910, opened 2023-04-17
Fix up logging message when discovery fails to find new service instances

trenner02:
The previous logging message contained no information about *which* service was failing:
```
Discovery returned zero available service instances. No previous service descriptors are available.
Discovery returned zero available service instances. No previous service descriptors are available.
Discovery returned zero available service instances. No previous service descriptors are available.
 ```

This PR just adds service and pool information to the logging messages so we know which service can't be discovered.


mphalpinpfpt:
+1, we have run into this issue too.

johngmyers (approved):

PR #911, opened 2023-04-28
Upgrade Jetty to 10.0.15

gdasbach:


gdasbach:
Some commented areas should be discussed and finalized before merging
* Removal of connection pool active requests in dump details
* Choice of excluding javax.servlet API (max 4.0.1), and keeping jetty's servlet API from toolchain (version 4.0.6)

http-client/src/main/java/com/proofpoint/http/client/jetty/JettyHttpClient.java
```diff
@@ -667,19 +661,21 @@ static List<JettyRequestListener> getRequestListenersForDestination(Destination
...
-        Queue<HttpExchange> httpExchanges = poolingHttpDestination.getHttpExchanges();
+        HttpDestination httpDestination = (HttpDestination) destination;
+        Queue<HttpExchange> httpExchanges = httpDestination.getHttpExchanges();
 
         List<org.eclipse.jetty.client.api.Request> requests = httpExchanges.stream()
                 .map(HttpExchange::getRequest)
                 .collect(Collectors.toList());
 
-        ((DuplexConnectionPool) poolingHttpDestination.getConnectionPool()).getActiveConnections().stream()
+        /* TODO: Need a better way to dump active requests due to getActiveConnections being removed
```
johngmyers, line 671: See Airlift. They drop a class into the eclipse package in order to export that private state.

http-server/src/main/java/com/proofpoint/http/server/HttpServer.java
```diff
@@ -236,7 +236,7 @@ private ServerConnector createHttpsServerConnector(HttpServerConfig config, Serv
 
         configuration.addCustomizer(new SecureRequestCustomizer());
 
-        SslContextFactory sslContextFactory = new SslContextFactory.Server();
+        SslContextFactory.Server sslContextFactory = new SslContextFactory.Server();
         sslContextFactory.setEndpointIdentificationAlgorithm("HTTPS");
         sslContextFactory.setKeyStorePath(config.getKeystorePath());
         sslContextFactory.setKeyStorePassword(config.getKeystorePassword());
```
johngmyers, line 242: We should remove the `if (!isJava8) {` test below as it will never be java8.

johngmyers:
I think we want to keep with the official javax.servlet API for now. We can switch to Jetty's servlet API when we go to Jetty 11 if need be. We should only do one servlet transition, as the one will be painful enough as is.

http-client/src/main/java/com/proofpoint/http/client/jetty/JettyHttpClient.java
gdasbach: Yeah, wasn't sure on whether we wanted to copy that behavior given that Jetty was explicitly hiding the implementation details, but easy enough to add

http-client/src/main/java/com/proofpoint/http/client/jetty/JettyHttpClient.java
```diff
@@ -180,20 +178,24 @@ public JettyHttpClient(
             sslContextFactory.setTrustAll(true);
         }
 
+        ClientConnector clientConnector = new ClientConnector();
+        clientConnector.setSslContextFactory(sslContextFactory);
+
         HttpClientTransport transport;
         if (config.isHttp2Enabled()) {
-            HTTP2Client client = new HTTP2Client();
+            HTTP2Client client = new HTTP2Client(clientConnector);
```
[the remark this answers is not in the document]
gdasbach: So airlift's experimental http/2 support excludes https by default, which is why this differs (and there is a strong assertion in the latest code to prevent secure connections for http2.  We look to have added https support for http2 connections, and added explicit unit tests around that support.  So really depends on whether we want to keep the https support for http2 or remove it, and the associated tests.

http-client/src/main/java/com/proofpoint/http/client/jetty/JettyHttpClient.java
gdasbach: Done

http-server/src/main/java/com/proofpoint/http/server/HttpServer.java
gdasbach: Done

gdasbach:
> I think we want to keep with the official javax.servlet API for now. We can switch to Jetty's servlet API when we go to Jetty 11 if need be. We should only do one servlet transition, as the one will be painful enough as is.

Re-added the javax.servlet-api, for version 4, and excluded the jetty toolchain implementation.  One consideration is that the jetty version supports JPMS, which the J2EE implementation won't support until version 6, I believe.  I don't think we're really using JPMS yet, so probably won't be a showstopper

johngmyers (changes requested):

http-client/src/main/java/com/proofpoint/http/client/jetty/JettyHttpClient.java
```diff
@@ -667,14 +661,14 @@ static List<JettyRequestListener> getRequestListenersForDestination(Destination
...
-        Queue<HttpExchange> httpExchanges = poolingHttpDestination.getHttpExchanges();
+        HttpDestination httpDestination = (HttpDestination) destination;
+        Queue<HttpExchange> httpExchanges = httpDestination.getHttpExchanges();
 
         List<org.eclipse.jetty.client.api.Request> requests = httpExchanges.stream()
                 .map(HttpExchange::getRequest)
                 .collect(Collectors.toList());
 
-        ((DuplexConnectionPool) poolingHttpDestination.getConnectionPool()).getActiveConnections().stream()
+        getActiveConnections((AbstractConnectionPool) httpDestination.getConnectionPool()).stream()
```
johngmyers, line 671: Need to `import org.eclipse.jetty.client.AbstractConnectionPool;`

johngmyers:
Please add an entry to CHANGES

http-client/src/main/java/com/proofpoint/http/client/jetty/JettyHttpClient.java
gdasbach: I thought I had pushed that yesterday.  Will be in the next push

johngmyers:
Please squash and change title to "Upgrade Jetty to 10.0.15"

gdasbach:
> Please squash and change title to "Upgrade Jetty to 10.0.15"

Done

johngmyers (approved):

PR #913, opened 2023-05-08
Upgrade Jersey to 2.34

gdasbach:
Jersey upgrade moves dependencies from javax.* to jakarta.* maven coordinates (javax.ws.rs-api -> jakarta.rs.rs-api, and  javax.annotation-api -> jakarta.annotation-api)

javax.validation:validation-api 2.0.1.Final re-introduced to provide compatibility with JEE 8 (for Jetty and Jersey), due to a conflict with Hibernate Validator which also uses jakarta.* coordinates but is based on JEE 9.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,7 +1,21 @@
+Platform 2.87
+
+* Jersey upgrade moves dependencies from javax.* to jakarta.* maven coordinates
+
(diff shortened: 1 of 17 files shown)


johngmyers:
Obsoleted by #915

PR #918, opened 2023-05-20
Replace airline with airline2

gdasbach:
Replace io.airlift:airline (0.4) with com.github.rvesse:airline (2.9.0) for compatibility with jakarta.inject


gdasbach:
Retracted due to duplicate PR https://github.com/proofpoint/platform/pull/917

PR #936, opened 2023-09-22
Upgrade logback to 1.4.11

johngmyers:


log/src/main/java/com/proofpoint/log/Logging.java
```diff
@@ -474,13 +474,7 @@ protected void subAppend(T event)
         private void flush()
         {
             try {
-                lock.lock();
```
johngmyers, line 477: `lock` is now private, so inaccessable. It doesn't look necessary to hold it in order to flush the output stream.

ehatch90 (approved):

PR #957, opened 2024-03-07
Allow the admin endpoint to have HSTS header

tomjshore:
This change is to allow the [HSTS headers ](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security#preloading_strict_transport_security) to be present on all admin endpoint.

Our security team was complaining that for a couple of admin endpoints that we added we needed this header. 

It is configured the same as normal by setting
jaxrs.hsts.max-age

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,10 @@
+Platform 3.18
+
+* Jaxrs
+
+  - When HSTS headers are enabled (by configuring "jaxrs.hsts.max-age"), we now attach the
+    Strict-Transport-Security header to responses from admin endpoints as well.
+
 Platform 3.17
 
 * Library Upgrades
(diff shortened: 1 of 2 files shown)


johngmyers (approved):

johngmyers (changes requested):
Please update the CHANGES file as well

CHANGES
```diff
@@ -1,3 +1,9 @@
+Platform 3.18
+
+* Jaxrs
+
+  - Allow HSTS filters to admin endpoints
```
johngmyers, line 5:
```suggestion
  - When HSTS headers are enabled (by configuring "jaxrs.hsts.max-age"), we now attach the
     Strict-Transport-Security header to responses from admin endpoints as well.
```

johngmyers (changes requested):

tomjshore:
All suggestions done.

johngmyers (approved):

PR #999, opened 2025-03-04
Use aarch64-compatible jnr-posix

paulroyal:
Bump jnr-posix to the final 3.0.x version to make launcher.jar compatible with aarch64 platforms.


johngmyers (changes requested):
Please update the CHANGES file as well.

